PR #10, opened 2018-02-13
Need mock logger info

chunfantasy:


mocks/coreMock.js
```diff
@@ -1,6 +1,6 @@
 // Allows jest to test files that import nrfconnect/core.
 
 module.exports = {
-    logger: {},
+    logger: () => { info: () => {} },
```
mrodem: The logger is an object, right? In that case, I think it should be:

```
logger: { info: () => {} }
```

You could also add `error`, `debug`, and `warn` while you are at it.

However, also try to see if you can avoid mocking. It is often a good idea to keep side-effects like logging away from (complex) logic. If you keep the logic in pure functions, then you don't have to mock anything when writing unit tests, and you also get some other nice benefits. See f.ex: https://hackernoon.com/javascript-and-functional-programming-pt-3-pure-functions-d572bb52e21c

chunfantasy:
Found a better solution.

PR #25, opened 2019-11-07
Switch from node-sass to dart-sass

datenreisender:
[node-sass](https://github.com/sass/node-sass) has a natively compiled portion, which leads to recurring
problems, e.g. with new node versions.

[dart-sass](https://sass-lang.com/dart-sass) (whose npm module is just called sass) on the other hand is
compiled to pure javascript and also the current reference
implementation.

Using dart-sass showed no major performance disadvantages and the
created output only differs in little things which should not lead to
different behaviour.

Please note, that this switch needs a change in the webpack
configuration. For the apps this is already done in this commmit, for
core this needs to be done when this version of devdep is used:

```diff
diff --git a/webpack.config.js b/webpack.config.js
index 60abac7..e04b300 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -47,10 +47,16 @@ module.exports = {
             exclude: /node_modules/,
         }, {
             test: /\.scss|\.css$/,
-            loaders: [
+            use: [
                 MiniCssExtractPlugin.loader,
-                require.resolve('css-loader'),
-                require.resolve('sass-loader'),
+                'css-loader',
+                {
+                    loader: 'sass-loader',
+                    options: {
+                        // eslint-disable-next-line global-require
+                        implementation: require('sass'),
+                    },
+                },
             ],
         }, {
             test: /\.(png|gif|ttf|eot|svg|woff(2)?)(\?[a-z0-9]+)?$/,
```


package.json
```diff
@@ -36,12 +36,12 @@
     "eslint-plugin-jsx-a11y": "6.2.3",
     "eslint-plugin-react": "7.14.2",
     "eslint-plugin-react-hooks": "1.6.1",
+    "fibers": "^4.0.2",
```
bencefr: Do we use fibers?

package.json
```diff
@@ -50,6 +50,7 @@
     "react-test-renderer": "16.8.6",
     "redux": "4.0.1",
     "rimraf": "2.6.3",
+    "sass": "^1.23.3",
```
bencefr: Should we pin the version?

package.json
datenreisender: [The documentation of `sass-loader` mentions `fibers`](https://github.com/webpack-contrib/sass-loader#implementation) and I had assumed that it speeds up compilation. But looking at it again I realised that I have no clue what they mean with “(a)synchronous compilation” and timing compilation with and without `fibers` I saw no difference, so I removed it again in b22a677c84e8.

package.json
datenreisender: Yes, I forgot that we usually ignore `package-lock.json` and instead fix the version in `package.json`. So I fixed that also in b22a677c8 and additionally added `save-exact = true` in my `˜/.npmrc` so this hopefully does not happen again.

With this, this PR is up for a review again.

PR #37, opened 2019-12-20
Update/device setup error handle

chunfantasy:
This PR is to let app to handle the no firmware defined error.

### DIFF
--- a/src/Device/deviceActions.js
+++ b/src/Device/deviceActions.js
@@ -292,7 +292,7 @@ export const selectAndSetupDevice = (
 
         await releaseCurrentDevice();
 
-        setupDevice(device, { promiseConfirm, promiseChoice, ...deviceSetup })
+        setupDevice(device, { promiseConfirm, promiseChoice, allowCustomDevice, ...deviceSetup })
             .then(preparedDevice => {
                 dispatch(startWatchingDevices(deviceListing, onDeviceDeselected));
                 dispatch(deviceSetupCompleteAction(preparedDevice));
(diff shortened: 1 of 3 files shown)


src/Device/deviceActions.js
```diff
@@ -289,19 +289,30 @@ export const selectAndSetupDevice = (
...
 
-        setupDevice(device, { promiseConfirm, promiseChoice, ...deviceSetup })
+        setupDevice(
+            device,
+            {
+                promiseConfirm,
+                promiseChoice,
+                allowCustomDevice,
+                ...deviceSetup,
+            },
```
datenreisender: Probably nicer is to extract this config object into a constant, in which you can then also inline `promiseConfirm` and `promiseChoice`. And if you even make `getDeviceSetupUserInput` curryable, then it looks just like this:

```js
const deviceSetupConfig = {
    promiseConfirm: getDeviceSetupUserInput(dispatch),
    promiseChoice: getDeviceSetupUserInput(dispatch),
    allowCustomDevice: false,
    ...deviceSetup,
};
setupDevice(device, deviceSetupConfig);
```

chunfantasy: Nice suggestion :+1:

src/Device/deviceActions.js
```diff
@@ -235,7 +235,7 @@ export const stopWatchingDevices = () => {
  * @param {Array<String>} [choices] The choices to display to the user (optional).
  * @returns {Promise<String>} Promise that resolves with the user input.
  */
-const getDeviceSetupUserInput = (dispatch, message, choices) => new Promise((resolve, reject) => {
+const getDeviceSetupUserInput = (message, choices) => dispatch => new Promise((resolve, reject) => {
```
datenreisender: This must be the other way around (you already bind the `dispatch` when calling it once and leave the other parameters to be bound later: 
```suggestion
const getDeviceSetupUserInput = dispatch => (message, choices) => new Promise((resolve, reject) => {
```

chunfantasy:
@datenreisender Sorry, I did not see your comment. But I actually had done the same things as you suggested, only Changelog.md is left. I will update it soon.

PR #43, opened 2020-03-12
Feature/user data

chunfantasy:
This PR is to support sending user data.

### DIFF
--- a/src/userData.js
+++ b/src/userData.js
@@ -52,12 +52,14 @@ const EventCategory = {
 
 /**
  * Initialize instance to send user data
- * @param {*} appName
+ * @param {*} appName the app's name e.g. Launcher
+ *
+ * @returns {void}
  */
 const init = async appName => {
     const networkInterfaces = await si.networkInterfaces();
-    const networkInterface = networkInterfaces.find(i =>
-        ((!i.virtual && i.mac) || i.iface === 'eth0'));
+    const networkInterface = networkInterfaces.find(i => (
+        (!i.virtual && i.mac) || i.iface === 'eth0'));
     const clientId = networkInterface
         ? shasum(networkInterface.ip4 + networkInterface.mac)
         : 'unknown';
@@ -80,9 +82,11 @@ const init = async appName => {
 
 /**
  * Send event
- * @param {EventCategory} category
- * @param {EventAction} action
- * @param {EventLabel} label
+ * @param {EventCategory} category launcher or apps
+ * @param {string} action action to collect
+ * @param {string} label details for an action
+ *
+ * @returns {void}
  */
 const sendEvent = (category, action, label) => reactGA.event({
     category,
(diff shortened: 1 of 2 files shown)


src/userData.js
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @returns {void}
+ */
+const init = async appName => {
+    const networkInterfaces = await si.networkInterfaces();
+    const networkInterface = networkInterfaces.find(i => (
+        (!i.virtual && i.mac) || i.iface === 'eth0'));
+    const clientId = networkInterface
+        ? shasum(networkInterface.ip4 + networkInterface.mac)
+        : 'unknown';
```
bencefr: How confident are you in this clientId? `networkInterface.ip4` is not checked and what if it has frequently changing IP address or only IPv6? I don't really have a suggestion though.

src/userData.js
```diff
@@ -0,0 +1,101 @@
...
+    LAUNCHER_CATEGORY: 'Launcher',
+    BLUETOOTH_LOW_ENERGY_CATEGORY: 'Bluetooth Low Energy',
+    PROGRAMMER_CATEGORY: 'Programmer',
+    LTE_LINK_MONITOR_CATEGORY: 'LTE Link Monitor',
+    TRACE_COLLECTOR_CATEGORY: 'Trace Collector',
+    POWER_PROFILER_CATEGORY: 'Power Profiler',
+    RSSI_CATEGORY: 'RSSI',
+    DIRECT_TEST_MODE_CATEGORY: 'Direct Test Mode',
+    TOOLCHAIN_MANAGER_CATEGORY: 'Toolchain Manager',
+};
```
bencefr: Are these categories to be updated as we go along, or could we just use the app name? Could we automatically figure it out from the caller's package.json?

src/userData.js
```diff
@@ -0,0 +1,101 @@
+/* Copyright (c) 2105 - 2019, Nordic Semiconductor ASA
```
bencefr: 2020

src/userData.js
chunfantasy: What we need here is actually a unique id to separate events from different users. We can have a look at the data later. If too many events are collected from a single user, then the logic here must be not good enough.

src/userData.js
chunfantasy: It is actually a simple string. Why I wanna put it here is just to unify the format. But since our app project name is unified, so we can use that name as the category. I will update this.

src/userData.js
```diff
@@ -0,0 +1,101 @@
...
+ * @returns {void}
+ */
+const init = async appName => {
+    const networkInterfaces = await si.networkInterfaces();
+    const networkInterface = networkInterfaces.find(i => (
+        (!i.virtual && i.mac) || i.iface === 'eth0'));
+    const clientId = networkInterface
+        ? shasum(networkInterface.ip4 + networkInterface.mac)
+        : 'unknown';
+    reactGA.initialize('UA-160398011-1', {
```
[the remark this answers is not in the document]
bencefr: Sorry, I meant to make a constant out of 'UA-160398011-1'.

PR #45, opened 2020-04-02
Switch to new serialport property path

datenreisender:
Part of [NCP-2996](https://projecttools.nordicsemi.no/jira/browse/NCP-2996).

Because this component is not used in old apps, we do not have to remain backward compatible.

Also switch to using the index as key, because the path/comName is not always unique.

### DIFF
--- a/src/Device/DeviceSelectorView.jsx
+++ b/src/Device/DeviceSelectorView.jsx
@@ -59,8 +59,7 @@ export default class DeviceSelector extends React.Component {
     static mapSerialPortsToListItems(device) {
         return Object.keys(device)
             .filter(key => key.startsWith('serialport'))
-            // eslint-disable-next-line react/no-array-index-key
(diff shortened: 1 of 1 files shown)


bencefr:
How is it possible to not have a unique path?  Using the index you can construct a string to avoid disabling the rule.

datenreisender:
> How is it possible to not have a unique path? 

I do not know why this happens, but I observed it in some cases, e.g. when opening the Link Monitor with a 91DK, then the last entry has the same path showing up twice: 

image

> Using the index you can construct a string to avoid disabling the rule.

I tried that, but the rule also complains about that. But just now I realised, that I can use something else: There is an parameter called `key` which I can use for the key (d'oh!). Just pushed a commit with that.

bencefr:
That is a bug in LinkMonitor, there should be only one serialport for the third device.

PR #50, opened 2020-05-18
New device selector

datenreisender:
This implements [NCP-2850](https://projecttools.nordicsemi.no/jira/browse/NCP-2850).

This branch is based on [the branch `new-design`](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/compare/new-design) and should also be merged back into it. The main change is 24eefd5, the other commits are just smaller changes around that. 

Some things are still missing, but the intention is to get the main building blocks in place now and iron out smaller things later:
* There is not yet an icon for unknown devices.
* The COM-port / serialport name still needs to be shortened when it is too long (which happens esp. on macOS)
* I want to add the PCA number to the details, since it is helpful sometimes
* The dialog when a device needs to be set up looks a bit dated (attached a screenshot). Maybe you have an idea for a new design.
* The RSSI app hangs if I decline to setup a device (but that seems already broken in the old app)
* I would like to make the About page responsive: Put a card below the others if it does not fit horizontally
* Hiding the sidepanel breaks the layout
* Some aspects that I would like to implement but maybe not now:
  * Renaming and favouriting devices is not implemented yet
  * Handling the keyboard focus is currently inconsistent: Some elements show if they have the focus, others don't. Would be good for a11y if all would show their focus.
  * In the design for the About page there is a button to show the release notes of the app which I did not implement yet.
  * When no devices are connected to the computer, instead of the device list I now show a short message pointing the user towards https://www.nordicsemi.com/Software-and-tools/Development-Kits . If an app is just for certain devices, then it would be even nicer if the message would point to those. (And also do that if there are devices connected but not the right ones)
  * When clicking on more device details, the design shows an expand animation that would be nice
  * It would be good to add a lint rule for apps that a repository location must be set in `package.json`, since we now use that in the about page


src/Device/DeviceSelector/DeviceList/Device.jsx
```diff
@@ -0,0 +1,102 @@
...
+    device: deviceShape.isRequired,
+};
+
+const additionalClassName = (moreVisible, isSelected) => {
+    if (moreVisible) return 'more-infos-visible';
+    if (isSelected) return 'selected-device';
+    return '';
+};
+
+const Device = ({ device, isSelected, doSelectDevice }) => {
```
bencefr, line 74: If I read this right currently you can expand the device to show the list of its serialports. The idea is that user can select serialports individually. Will such functionality be part of this PR? I think optionally extending the result with an index or serialport.path would be perfect.

src/Device/deviceInfo/deviceInfo.jsx
```diff
@@ -110,14 +155,38 @@ const deviceInfo = pcaNumber => ({
     pca20035: {
```
bencefr, line 155: I suggest to use all caps 'PCA20035' and sanitise the pca key, better safe than sorry when a firmware update comes.

src/Device/deviceInfo/deviceInfo.jsx
```diff
@@ -34,10 +34,23 @@
...
+import nrf51logoWhite from './nRF51-Series-logo_white.svg';
+import nrf51logoBlue from './nRF51-Series-logo_blue.svg';
+import nrf52logoWhite from './nRF52-Series-logo_white.svg';
+import nrf52logoBlue from './nRF52-Series-logo_blue.svg';
+import nrf53logoWhite from './nRF53-Series-logo_white.svg';
+import nrf53logoBlue from './nRF53-Series-logo_blue.svg';
+import nrf91logoWhite from './nRF91-Series-logo_white.svg';
+import nrf91logoBlue from './nRF91-Series-logo_blue.svg';
+
 const deviceInfo = pcaNumber => ({
```
bencefr, line 46: Looking at the device-lister mapping PCA10068 and PCA10115 are missing.

src/Device/DeviceSelector/DeviceList/Device.jsx
datenreisender: I don't understand this aspect about serial ports yet and so was too ignorant to implement it. Probably would help me, if I can see it in action: In which app can users select serial ports individually?

src/Device/deviceInfo/deviceInfo.jsx
datenreisender: Ah, yes, I remember that you already commented that on the other PR which I did not yet get around to implement. Implemented now here in 261b9bd.

src/Device/deviceInfo/deviceInfo.jsx
datenreisender: I see the entries in https://github.com/NordicSemiconductor/nrf-device-lister-js/blob/facb4aa9e/src/util/board-versions.js but I have no clue what they are. Can you tell me what they are? Or is this something we have not (yet?) released and thus should not (yet) include with details here?

bencefr: These are all the PCA numbers we encountered so far. There are boards with all these numbers, maybe some of them are not available to purchase but they exist.

src/Device/DeviceSelector/DeviceList/Device.jsx
bencefr: Link monitor and Trace collector. All 91 gadgets have 2 or 3 serial ports and since it's just a guess which one is the correct, the user should be able to select if auto-selection does not work. See "Auto device/port filter" functionality of these apps.

datenreisender: Ah, yes, let us care about them later and also discuss a good UI for that case with Ketil when we update one of these apps.

bencefr: You are right about that this is not yet needed, however we have an upcoming 91 app that will most likely require the same functionality. Agreed, not needed in this PR though.

src/Device/deviceInfo/deviceInfo.jsx
datenreisender: @chunfantasy: I think you added PCA10068 and PCA10115 to the device-lister. Do you have any info on what the boards for those numbers are called and which chip is on them? If not (or if we do not want to disclose it) then I suggest we leave them out. In that case the device selector will show the PCA number instead of the name (before falling back to “Unknown”): https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/261b9bd8f81583ce9dbe252e0a3703ca40088a0a/src/Device/DeviceSelector/BasicDeviceInfo.jsx#L47

bencefr: As there is a fallback I'd vote for not holding back the PR for this.

PR #55, opened 2020-06-16
Feature/nicknames

sunnivablock:
We have created functionality for creating nickname and favorite devices. The values are saved in persistent store. There is however a error with three of the tests that we have been unable to solve and need help with solving.


src/Device/deviceReducer.js
```diff
@@ -60,8 +63,13 @@ const initialState = {
 
 export default (state = initialState, action) => {
     switch (action.type) {
-        case DEVICES_DETECTED:
-            return { ...state, devices: [...action.devices] };
+        case DEVICES_DETECTED: {
+            const devices = action.devices.map(d => ({
+                ...d,
+                favorite: getIsFavoriteDevice(d.serialNumber),
```
bencefr, line 69: I'd suggest to add `nickname` to the device as well, it will help to use it for sorting and to avoid using _getDeviceNickname()_.
```
    nickname: getDeviceNickname(d.serialNumber),
```

src/Device/DeviceSelector/DeviceList/DeviceList.jsx
```diff
@@ -61,10 +61,11 @@ const DeviceList = ({ devices, doSelectDevice }) => {
     const selectedSerialNumber = useSelector(selectedSerialNumberSelector);
 
     if (devices.length === 0) return <NoDevicesConnected />;
+    const sorted = devices.sort((a, b) => (b.favorite - a.favorite));
```
bencefr, line 64: This way of sorting has random behaviour between two favourited or unfavourited device.
If you've added the _nickname_ to the reducer as I suggested in another comment, it makes this sorting easier:
```js
const sorted = devices.sort((a, b) => {
  const f = (b.favorite - a.favorite) + 1;
  return (`${f}${a.nickname}` > `${f}${b.nickname}` ? 1 : -1);
};
```
Note the comparison between booleans gives a value between -1 and 1, thus adding 1 makes _f_ between 0 and 2...

src/Device/deviceReducer.js
sunnivablock: We have implemented this in the new version that we will push soon, but it wont update live so we are not able to use it

bencefr: I suspected that since there's no redux workflow to change the nickname.

src/Device/DeviceSelector/DeviceList/DeviceList.jsx
sunnivablock: This sorting does not work for me at all. I can end up with a starred item at the bottom, does it work for you?

datenreisender: @bencefr: Even though the spec does not require it, [in most engines nowadays `Array.prototype.sort` is stable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Browser_compatibility), so the order should not be random. Or is the order in the incoming array already random?

In case we still want to sort them by name: Using automatic conversion of booleans into numbers to do arithmetics with them? That is a hackish way to solve things that I always have to look up later so that I understand how it works. :-D

I would rather implement it this way:

```js
const sorted = devices.sort((a, b) => {
  if (a.favorite !== b.favorite) {
  	return a.favorite ? -1 : 1;
  }
  
  return a.nickname < b.nickname ? -1 : 1;
};
```

If we are sure that the `nickname` property is always a defined string, then I would replace the last line with `return a.nickname.localCompare(b.nickname)`.

bencefr: It cannot be not random if the sort function does not take the name into account. I don't know what stable wants to mean by the spec, but this is random. Indeed your function does look more readable :)

datenreisender: A sorting algorithm is stable, if it keeps the original order for two elements that are considered equal by the order criterion. In this case: If we only sort based on the favourite flag, the order within the favourited and not favourited devices should stay the same, if we ignore other criteria.

bencefr: For me that is still random, since it is not humanly controlled what is the criteria. It may be memory addresses of the objects it compares, time of their creation or anything. It can be stable if the order is always the same, but if there's no human logic to it, I'll still won't find the item in the list.

src/About/about.scss
```diff
@@ -20,7 +20,7 @@ $section-gap: 2ex;
         padding-left: 2ex;
         padding-right: 2ex;
 
-        padding-top: 1.2ex;
+        padding-top: 1.3ex;
```
datenreisender, line 23: Why did you change this? Since you did not work on the about pane, I assume this was just done by mistake, so please revert it.

PR #58, opened 2020-07-13
Rename with inline input

datenreisender:
With these changes, clicking on “Rename device” (or on the device name) lets users rename the device right in the device name.

### DIFF
--- a/src/Device/DeviceSelector/BasicDeviceInfo.jsx
+++ b/src/Device/DeviceSelector/BasicDeviceInfo.jsx
@@ -35,19 +35,46 @@
  */
 
 import React from 'react';
-import { bool, node } from 'prop-types';
+import {
+    bool, func, node, shape,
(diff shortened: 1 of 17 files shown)


src/Device/DeviceSelector/BasicDeviceInfo.jsx
```diff
@@ -72,6 +101,8 @@ const BasicDeviceInfo = ({ device, whiteBackground, additionalToggle }) => (
 );
 BasicDeviceInfo.propTypes = {
     device: deviceShape.isRequired,
+    // eslint-disable-next-line react/require-default-props
```
chunfantasy, line 104: If this rule need to be disabled, shall we add it in eslint config?

datenreisender: Yes, I will prepare a PR for it.

PR #64, opened 2020-07-21
Sidepanel redesign updates

datenreisender:
Part of [NCP-3117](https://projecttools.nordicsemi.no/jira/browse/NCP-3117), this belongs together with NordicSemiconductor/pc-nrfconnect-rssi#57.

* Sliders: On hover the handle enlarges and changes it's colour. Size change should stay but no colour change.
* Incorporate some code, that was previously in the RSSI app

### DIFF
--- a/src/bleChannels.js
+++ b/src/bleChannels.js
@@ -34,30 +34,14 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import * as ErrorDialogActions from './ErrorDialog/errorDialogActions';
+export const bleChannels = [
+    37, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 38, 11, 12, 13, 14, 15, 16, 17, 18,
+    19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 39,
+];
 
-export { ErrorDialogActions };
+export const bleChannelRange = {
+    lower: Math.min(...bleChannels),
+    upper: Math.max(...bleChannels),
+};
 
-export { default as App } from './App/App';
-export { default as Logo } from './Logo/Logo';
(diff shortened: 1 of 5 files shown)


src/bleChannels.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const bleChannels = [
+    37, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 38, 11, 12, 13, 14, 15, 16, 17, 18,
+    19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 39,
+];
+
+export const bleChannelRange = {
+    lower: Math.min(...bleChannels),
+    upper: Math.max(...bleChannels),
+};
```
bencefr, line 45: How do you plan to use this? For me _lower_ and _upper_ sounds like a function, however their values are the min and max.
You also could just assign the min and the max to the array itself:
```js
bleChannels.min = 0;
bleChannels.max = 39;
```

datenreisender: I am currently using this in two places: In the RSSI app in [the reducer]( https://github.com/NordicSemiconductor/pc-nrfconnect-rssi/blob/7a10e34a5b04144df04bdf20cd10d93ce54aea3d/src/reducer.js#L52) and in [the component to select the displayed channel range](https://github.com/NordicSemiconductor/pc-nrfconnect-rssi/blob/7a10e34a5b04144df04bdf20cd10d93ce54aea3d/src/SidePanel/ChannelRange.jsx#L60-L73).

I also struggle with the names. To me `min` and `max` more sound like functions (maybe because of `Math.min/max`) so I started using `lower` and `upper` in several places but also thought about `start` and `end` occasionally. Maybe I will think about that again and see whether it looks better when I depart from `lower/upper` in the RSSI app.

Usually I avoid putting additional properties on Arrays, being afraid that they mess up other code unexpectedly. But I could use `Object.defineProperty` which would at least not make `min` and `max` (and `isAdvertisementChannel`) enumerable properties. I like that idea and will do so in another PR.

PR #69, opened 2020-07-31
Render all panes

bencefr:
This PR is important for any app that maintains objects/state/logic which is created by the components when they get rendered in the main view. Since the About pane is automatically added, even single pane apps are effected by this flaw.

Potential improvement would be to animate the change between panes, maybe to use some-kind of a Carousel component.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+## Unreleased
+### Changed
+- Render all panes of the App, but only the current is visible
+
 ## Version 4.8.4
 ### Added
 - Favorite and nickname in device selector for new design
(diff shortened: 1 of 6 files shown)


datenreisender:
Usually, when a component is not shown I also prefer to not render it. 

Preventing unnecessary re-renders because an invisible component is changed is just one reason for that. Another is that it usually simplifies the logic within a component if it can rely on the assumption that it is only rendered if the system is in a fitting state and that things like lifecycle function are called at appropriate times. We could reach some of these goals through other ways, e.g. by telling a main pane whether it is currently visible or not, but I only would add this, if we see a concrete need for it in one of our apps.

But I assume that you have a certain use case why you would like to always render all panes. Which use case is that?

The animation you mention as possible improvement is currently not planned, AFAIK.

bencefr:
Basically all apps that render a component that depends on initialising stuff in `useEffect(..., [])` will be effected. Even if there used to be a single main view, now the About pane will make them a multi pane app.
Since the introduction of hooks more and more app started to introduce local state and logic in those components, because it was easier to implement avoiding the full redux path. If we force this new requirement, all these will have to be refactored which be quite some headache.

bencefr:
Yes, animation is not planned, it's a suggestion to make it more pleasing, like other animation you introduced lately.

PR #71, opened 2020-08-11
Update/usage data dialog

chunfantasy:
This PR is to update confirmation dialog style for usage data dialog in launcher.


Changelog.md
```diff
@@ -3,6 +3,7 @@
 - The remote property on the electron mock was noch mocked correctly
 ### Changed
 - Render all panes of the App, but only the current is visible
+- Updated confirmation dialog style
```
datenreisender, line 6: 4.8.5 is already tagged, so this changelog entry should go under a new heading (either “Unreleased” or give it a new number like 4.8.6 and also bump the number in the `package.json`)

src/Dialog/ConfirmationDialog.jsx
```diff
@@ -34,13 +34,20 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import React from 'react';
+import './confirmation-dialog.scss';
```
datenreisender, line 37: In other files we put the SCSS imports always last. Let us stick to that convention.

(We could also change the convention to always put the SCSS imports first.)

src/Dialog/confirmation-dialog.scss
```diff
@@ -0,0 +1,10 @@
+@import "../variables";
+
+.modal {
+    background-color: $modal-backdrop-bg;
```
datenreisender, line 4: Instead of manually setting the background colour here, better change the backdrop value to `true` here (for confirmation dialogs that are not in progress): 
https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/85722f36bbccba24dff231e9347c04f5ff695a93/src/Dialog/ConfirmationDialog.jsx#L85

src/Dialog/confirmation-dialog.scss
```diff
@@ -0,0 +1,10 @@
...
+}
+
+.modal-content {
+    border-radius: 0px !important;
+    border-color: $modal-content-border-color;
```
datenreisender, line 9: My impression is, that this gets overridden anyhow, so you can remove it.

src/Dialog/confirmation-dialog.scss
```diff
@@ -0,0 +1,10 @@
+@import "../variables";
+
+.modal {
+    background-color: $modal-backdrop-bg;
+}
+
+.modal-content {
+    border-radius: 0px !important;
```
datenreisender, line 8: This (like the whole file) will effect not just confirmation dialogs but _every_ modal. Is this your intention? 

If yes, then put it in a more central place to indicate this (currently, I suggest `shared.scss` for it), if not, then add a specific class name to the modal of the confirmation dialog and also use it here (which would also make it unnecessary to use `!important`, since it would increase the specificity of the selector).

src/Dialog/ConfirmationDialog.jsx
```diff
@@ -72,9 +79,16 @@ const ConfirmationDialog = ({
...
+    <Modal
+        show={isVisible}
+        onHide={onCancel}
+        backdrop={isInProgress ? 'static' : false}
+        size="lg"
+    >
         <Modal.Header closeButton={!isInProgress}>
-            <Modal.Title data-testid="title">{title}</Modal.Title>
+            <Modal.Title data-testid="title">
+                <h3>{title}</h3>
```
datenreisender, line 90: This change (and also the line https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/85722f36bbccba24dff231e9347c04f5ff695a93/src/Dialog/ConfirmationDialog.jsx#L110) do change other existing dialogs. Is this ok?

E.g. the version problem dialog before looked like this:

_Users_ms_Desktop_nRF%20Connect%203 3 0 app_Contents_Resources_app asar_resources_launcher html

and it is now changed to this:

_Users_ms_w_pc-nrfconnect-launcher_resources_launcher html

src/variables.scss
```diff
@@ -657,7 +657,7 @@ $modal-content-border-color:                   rgba(0,0,0,.2);
 $modal-content-fallback-border-color:          #999;
 
 //** Modal backdrop background color
-$modal-backdrop-bg:           #000;
+$modal-backdrop-bg:           rgba(0,0,0,.5);
```
datenreisender, line 660: The opacity is applied automatically through the next settings for normal backdrops. So you can revert this if you set a backdrop as described above.

Changelog.md
chunfantasy: ok

src/Dialog/ConfirmationDialog.jsx
chunfantasy: image
I just use the `sort-imports` extension with its default settings to sort all imports. It's very convenient. I would prefer to use it to make the order consistent. Do you think if that is ok? I can update the other files with the extension if it is ok. Or I can just move the single line.

src/Dialog/confirmation-dialog.scss
chunfantasy: Good point, I will update.

src/Dialog/confirmation-dialog.scss
chunfantasy: Removed.

src/Dialog/confirmation-dialog.scss
chunfantasy: Yes, it is my intention, I will move it to shared.scss.

src/Dialog/ConfirmationDialog.jsx
chunfantasy: It is ok. I mentioned it to Ketil and he agreed.

src/variables.scss
chunfantasy: Reverted.

src/Dialog/ConfirmationDialog.jsx
chunfantasy: ok, since I removed .scss file, the import is not needed any more.

datenreisender: The extension looks nice; even better would be if we use an ESLint plugin for it, since then it would be checked on the build server and everybody could use it regardless of the IDE. [eslint-plugin-simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) looks like such a plugin, in case you want to look further into that.

PR #74, opened 2020-08-19
Style enhancements

datenreisender:


src/App/shared.scss
```diff
@@ -30,7 +32,21 @@ body {
...
+    border: 0.39em solid;
+}
+
 // used when tabbing and space
 :focus {
     outline-color: $accent !important;
+    outline-width: 3px;
+}
+:focus:active {
+    outline-style: none;
```
datenreisender, line 51: @bencefr Why did you alter the two settings for `outline-width` and `outline-style` here? I could not find any visual changes resulting from this.

bencefr: Outline is drawn on elements which are in focus. It's ugly, Ketil does not like it. You can get focus on an element by clicking on it, or by keyboard tabbing. When using the keyboard, it is essential to know where the focus is, but when clicking on a button, you don't want that ugly outline to remain there.
The difference comes from when you use the tab, the element gets focus, but is not active yet, (until you also hit space). If you use the mouse, it is both focused and active.

datenreisender: Ah, yes, focus is a double-edged sword. In order not to leave the outline when clicking on a pseudo button, I called `blur()` in https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/f5c51b1725a693c0a2602c6323140019c907dfa8/src/PseudoButton/PseudoButton.jsx#L53, but maybe with this change that is not necessary anymore. I will check that.

Thanks for the explanation.

bencefr (approved):
👐

PR #76, opened 2020-08-24
Feature/toggle

chunfantasy:
This PR is to add Toggle component.

image

``` html
<div style={{padding: '50px'}}>
        <Toggle variant="primary" />
        <Toggle variant="secondary" />
        <Toggle variant="primary" disabled />
        <Toggle variant="secondary" disabled />
        <Toggle variant="primary">Primary</Toggle>
        <Toggle variant="secondary">Secondary</Toggle>
        <Toggle
            variant="secondary"
            barColor="#b0bec5"
            barColorToggled="#b0bec5"
            handleColor="#78909c"
        >
            Customized
        </Toggle>
        <Toggle
            variant="secondary"
            barColor="#b0bec5"
            barColorToggled="red"
            handleColor="#78909c"
        >
            Customized
        </Toggle>
        <Toggle
            variant="primary"
            width="400px"
        >
            Primary with width 400px
        </Toggle>
        <Toggle
            variant="secondary"
            width="400px"
        >
            Secondary with width 400px
        </Toggle>
        <Toggle
            variant="primary"
            labelRight
            width="400px"
        >
            Label right
        </Toggle>
        <Toggle
            variant="secondary"
            labelRight
            width="400px"
        >
            Label right right
        </Toggle>
    </div>
```

### DIFF
--- a/src/Toggle/Toggle.jsx
+++ b/src/Toggle/Toggle.jsx
@@ -1,0 +1,81 @@
+/* Copyright (c) 2015 - 2017, Nordic Semiconductor ASA
+ *
+ * All rights reserved.
+ *
+ * Use in source and binary forms, redistribution in binary form only, with
+ * or without modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions in binary form, except as embedded into a Nordic
+ *    Semiconductor ASA integrated circuit in a product or a software update for
+ *    such product, must reproduce the above copyright notice, this list of
+ *    conditions and the following disclaimer in the documentation and/or other
+ *    materials provided with the distribution.
+ *
+ * 2. Neither the name of Nordic Semiconductor ASA nor the names of its
+ *    contributors may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ *
+ * 3. This software, with or without modification, must only be used with a Nordic
+ *    Semiconductor ASA integrated circuit.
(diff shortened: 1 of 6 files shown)


src/Toggle/Toggle.jsx
```diff
@@ -0,0 +1,146 @@
...
+    handleColor,
+    handleColorToggled,
+    width,
+    disabled = false,
+    children,
+}) => {
+    const [toggled, setToggled] = useState(isToggled || false);
+    const isPrimary = variant === 'primary';
+    const isSecondary = variant === 'secondary';
+    const elementId = id || `Toggle-${Math.random().toString(36)}`;
```
bencefr, line 62: What effect does this have if _id_ is undefined? Would it assign a new random id at every render?
I think it's better to require _id_ to be defined.

bencefr (commented):
Great and simple implementation!

src/Toggle/Toggle.jsx
chunfantasy: Without `id` wont work and will lead to eslint error.
With the same `id` will lead to triggering different Toggles wit the same id.
I am a bit uncertain here. I just changed id from require to optional. Because I think it can make less attributes to set for the component.
And I understand your point, too.
So I can revert back. Both solutions are fine to me.

bencefr (approved):

PR #79, opened 2020-08-26
Fix the Toggle components state in case it's controlled from outside

bencefr:
Should the state of the Toggle component depend on a value that has changed outside of the Toggle itself would end up in a mismatch between the internal state and the external value.
This PR is meant to fix this case.
Additionally the onToggle callback now provides the newly set value.


chunfantasy (approved):
Maybe we can also remove the internal state. I dont see any points to have it instead of testing locally.

PR #81, opened 2020-08-27
Finer slider

bencefr:
- Added optional chars prop to define number of characters of NumberInput
- Added optional decimals prop to range of Slider

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+## Version 4.8.13
+### Added
+- Added optional chars prop to define number of characters of NumberInput
+- Added optional decimals prop to range of Slider
+
 ## Version 4.8.12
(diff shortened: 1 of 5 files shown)


datenreisender (approved):
I did not know about `toFixed` before. Nice!

It seems a bit strange to me, that `decimals` is part of the range. To me it seems to be something different so I would also pass it in as a separate parameter but if you want to include it, that is also fine.

You could also use the decimals for calculating the possible character length, but keeping it simple by specifying it explicitly also has advantages.

bencefr:
I didn't want to change the props or propagate them so deep. I'd maybe call _range_ e.g. _options_ now, but this will rarely be used, and is also backwards compatible.

PR #85, opened 2020-09-03
Add TypeScript and Prettier support

bencefr:


config/eslintrc.json
```diff
@@ -1,8 +1,11 @@
 {
-    "parser": "babel-eslint",
+    "parser": "@typescript-eslint/parser",
     "extends": [
         "eslint:recommended",
-        "airbnb"
+        "airbnb",
+        "plugin:@typescript-eslint/recommended",
+        "prettier",
+        "prettier/@typescript-eslint"
```
datenreisender, line 8: I have seen in https://github.com/prettier/eslint-config-prettier that there also is a `prettier/react`. No idea what that changes and whether that also might be helpful.

config/eslintrc.json
```diff
@@ -63,18 +63,39 @@
         ],
         "import/extensions": ["off"],
         "no-undef": 1,
-        "no-unused-vars": 1
+        "no-unused-vars": "off"
```
datenreisender, line 66: You did add or change quite a lot of settings here (like this one or above `camelcase` or `no-param-reassign`) where it is hard for me to understand why you changed them, but I assume they were necessary.

config/webpack.config.js
```diff
@@ -1,10 +1,13 @@
...
 const path = require('path');
 const fs = require('fs');
-const dependencies = require(path.join(
+
+const { dependencies } = require(path.join(
     process.cwd(),
-    'package.json',
-)).dependencies;
+    'node_modules',
+    'pc-nrfconnect-shared',
```
datenreisender, line 8: Is this really correct? I know that there were some changes to this on the master branch and I fear this might be a merge gone wrong.

index.d.ts
```diff
@@ -0,0 +1,398 @@
+declare module 'pc-nrfconnect-shared' {
+    import { Reducer, AnyAction } from 'redux';
+    import React from 'react';
+    import winston from 'winston';
+
+    // State
+
+    interface NrfConnectState<AppState> {
```
datenreisender, line 8: Is the idea to keep these types in a separate file and not convert the normal JavaScript files to TypeScript, so that the types are defined (and used!) were they originate?

datenreisender (commented):
@deerob4: I think it would have been nicer if you would have done two different branches, one for adding Prettier and another one (maybe based on the previous) for adding TypeScript. But it is fine if you do not want to do the effort now anymore.


I am not sure whether I misunderstand what you mean with “Add Prettier support”. My expectation was, and maybe I am wrong here, that with this PR it is possible to either do the formatting with Prettier or to check the formatting with Prettier (or both) for all projects that use `pc-nrfconnect-shared` as well as for `pc-nrfconnect-shared` itself. 

But to test it [I introduced some formatting errors in a branch `demonstrate_formatting_errors`](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/compare/merge/ts-support...demonstrate_formatting_errors) and e.g. `npm run lint` does not give me any errors (it does, if I add the same errors on the master branch).

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "prettier.configPath": "./config/.prettierrc.json",
+    "eslint.options": {
+        "configFile": "./config/eslintrc.json"
+    }
+}
```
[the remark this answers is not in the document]
deerob4: You're right, this was an errant file that shouldn't have been added to Git in the first place. I use that file in`shared` since my global setting (pointing to the individual app's `node_modules/pc-nrfconnect-shared/config` of course doesn't work in `shared`.

Changelog.md
```diff
@@ -31,6 +36,8 @@
 ## Version 4.8.8
 ### Fixed
 - Fixed installation failure in local apps
+### Updated
+- Updated several design elements for the release of the new RSSI app
```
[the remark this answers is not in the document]
deerob4: Yep, something happened in the rebase :)

config/.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+    "singleQuote": true,
+    "tabWidth": 4,
+    "trailingComma": "es5",
+    "jsxBracketSameLine": false,
+    "arrowParens": "avoid",
+    "printWidth": 80
+}
```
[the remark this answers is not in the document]
deerob4: Yes, I have renamed that file to `prettier.config.js`. I don't really like the style either, but I thought it best to just mimic what was already there.

config/babel.config.js
```diff
@@ -1,15 +1,17 @@
...
+        plugins: [
+            'add-module-exports',
+            '@babel/plugin-proposal-class-properties',
+            '@babel/plugin-transform-destructuring',
+            '@babel/plugin-transform-modules-commonjs',
+            '@babel/plugin-transform-parameters',
+            '@babel/plugin-transform-spread',
+            '@babel/plugin-proposal-object-rest-spread',
+            '@babel/plugin-proposal-optional-chaining',
+            '@babel/plugin-syntax-dynamic-import',
```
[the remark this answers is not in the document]
deerob4: Just forgot to remove it :)

config/webpack.config.js
deerob4: This is correct, though it wasn't my addition. Chun added it when we had some problems getting the apps to build with the new config, a file resolution error somewhere. Somehow this solves that issue.

config/eslintrc.json
deerob4: I can't see what that changes either really; Prettier is working fine with React as it is so I don't think we need to add anything else.

config/eslintrc.json
deerob4: Yes, they were all necessary. In the case of the `no-unused-vars` one, the TypeScript ESLint plugin contains a duplicate rule of its own. Some also interfered with TypeScript conventions. And a couple didn't seem to make sense and were only included because they were pulled in by the recommended or Airbnb plugin. For example, I had a class with some twelve or so properties. Requiring a line between all of them made it harder to group them, reducing readability.

index.d.ts
deerob4: I think long-term we should move the original files to TypeScript, in which case this file would gradually become unnecessary. For now though, this was a nice way to put all the types in one place, without having to touch every file in the project. It's the approach used by most JavaScript libraries that need to be interfaced with from TypeScript.

chunfantasy:
I am just wondering, is it better to use merge than to use rebase in this case?

datenreisender:
Currently this PR uses [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) which turns _off_ ESLint rules that might conflict with Prettier.

How about if we additionally use [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier), which makes ESLint check (or correct) the formatting according to Prettier?

Alternatives might be to use something like [`prettier-eslint`](https://github.com/prettier/prettier-eslint) or run `prettier` manually but I think for now it is easiest to integrate it into ESLint with `eslint-plugin-prettier`.

.vscode/settings.json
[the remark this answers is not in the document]
datenreisender: Even though it is good that you removed it here, should we maybe describe how users can configure it in VS Code if they, wish to like https://github.com/NordicSemiconductor/pc-nrfconnect-docs/blob/8f48a0250e71bb73c4b1af4f7eb0608f47e26a4e/app_development.md#using-eslint-in-an-ide?

config/webpack.config.js
datenreisender: Chun did change this in https://github.com/NordicSemiconductor/pc-nrfconnect-shared/commit/d0058090606ff8559731fefa83d8f340106ceba3#diff-a58d55bdb5770c78ad512f8e91f8d051 but he changed it again in https://github.com/NordicSemiconductor/pc-nrfconnect-shared/commit/8b77d7afc373f8f185e450ea83fa3e7aeea22563#diff-a58d55bdb5770c78ad512f8e91f8d051 because we had problems with installing local apps. So the current version in `master` looks like this and yours is different, so please check this again: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/184a1054e30b5464a8b070e7ada02100b4a9a33b/config/webpack.config.js#L4-L7

deerob4:
We've now got Prettier and ESLIint working together, using that plugin. ESLint's standard formatting rules have been disabled, with all of those checks delegated to Prettier. I also added the `prettier/react` rules, so some conflicts have been fixed.

Like I mentioned in the standup, this change will require the following key to be added to the `package.json` of all the apps:

`{ "prettier": "./node_modules/pc-nrfconnect-shared/config/prettier.config.js" }`

This is because `eslint-plugin-prettier` doesn't allow the configuration file to be set "globally", as per [this issue](https://github.com/prettier/eslint-plugin-prettier/issues/246#issuecomment-555181557). Once this is merged, I will go through the apps and create pull requests for those changes.

datenreisender:
Nice! As an alternative (or additionally) to going through all apps yourselves, you could also add a subsection “ Steps to upgrade when using this package” to the changelog like I also propose in PR #86. There you could explain that apps using this version should add that `prettier` entry to `package.json` and that the linting will probably fail but can usually be fixed by running `npm run lint -- --fix`.

deerob4:
Good idea, added that!

datenreisender (approved):
Thanks for the effort! I am looking forward to more nicely formatted files and utilising TypeScript more.

PR #87, opened 2020-09-08
Enhance device selector

datenreisender:
Three small changes to the device selector, more specifically to the device details:
- Add PCA number
- Remove the wrong line below the port names
- Better alignment for the port names: Port names were centered, which looked okay for long names but out of place for shorter names. If they are short enough, they are now left aligned with the other text
above (device name, serial number, PCA number). If the port names are so
long that they would run to the right out of the device area, then they
port names are moved to the left just enough to fit.

How a device looked before:
Before

How it looks now:
After

How it looks now with short port names (before they were centred):
After with short port names

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+## Version 4.8.17
+### Updated
+- New design, device selector: Add PCA number, better alignment for port names,
+  removed wrong line below port names
+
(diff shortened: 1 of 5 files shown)


bencefr (approved):
LGTM

PR #94, opened 2020-09-16
Use focus-visible polyfill to separate keyboard focus from pointer focus

bencefr:
This PR removes outline and box-shadow of focused elements if it was focused by pointer, but not when focus was received by keyboard.


chunfantasy (approved):
LGTM

PR #99, opened 2020-09-24
Add missing Redux action types

deerob4:
There were a few action types and constants that were missing from the `NrfConnectAction` union; this PR adds them.

For reference, the type can be used like this:

```ts
// actions.ts

import { NrfConnectAction } from 'pc-nrfconnect-shared';

type AppAction =
    | DeviceAction
    | SettingsAction
    | NrfConnectAction // <-- just add it to the end of the union type
```

This way, reducers and the `dispatch` can react to the nRF Connect actions as well, with full type information.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,6 +41,17 @@ declare module 'pc-nrfconnect-shared' {
         LOG_ADD_ENTRIES = 'LOG_ADD_ENTRIES',
         LOG_CLEAR_ENTRIES = 'LOG_CLEAR_ENTRIES',
         LOG_TOGGLE_AUTOSCROLL = 'LOG_TOGGLE_AUTOSCROLL',
+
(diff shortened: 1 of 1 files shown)


datenreisender:
Why do you want to expose all these action and action types? Do you currently need them?

I would prefer to keep them internal implementation details of `shared` instead of adding them to the API. But if there are cases were an app wants to e.g. dispatch a `SelectDevice` action or react on it, I would like to understand it better. So far we have added callbacks on the `DeviceSelector` for all the moments I saw, when apps need to hook into certain events.

deerob4:
No, I don't need them, at the moment at least. Mostly I added them for completeness' sake. I guess it's two different ways of thinking about the API we expose, between explicit callbacks and letting consumers listen to the actions themselves.

Since we are already using callbacks, and it's a bit cleaner, I'm happy to close this in favour of continuing with the current approach.

PR #100, opened 2020-09-29
Added optional range.decimals to be validated for NumberInlineInput

bencefr:


datenreisender (approved):
Looks good, nice catch that you also updated `index.d.ts`. 

I am undecided whether “4.9.6” is the right version number. Since you rather added a small feature than fixed one, I would call it “4.10.0” but going with “4.9.6” is also ok.

PR #101, opened 2020-10-01
Enable showing info for USB connected devices

datenreisender:
Enable showing an own icon and links on the About pane for USB connected devices.

This PR does convert a first file from JavaScript to TypeScript: `deviceInfo`. While most is already prepared for that, the `webpack.config.js` in the launcher still needs two adjustments as described in the `Changelog.md`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,19 @@
+## 4.10.0
+### Added
+- Enable showing an own icon and links on the About pane for USB connected
+  devices.
+### Steps to upgrade when using this package
+- This version does introduce a first file in `shared` that is converted from
+  JavaScript to TypeScript: `deviceInfo`. While most is already prepared for
+  that, the `webpack.config.js` in the launcher still needs two adjustments to
+  work with this: The lines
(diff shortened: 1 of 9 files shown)


deerob4:
Very nice! Perhaps the only change I would make is in the function signature of `devicesByPca` at `deviceInfo.ts:80`. Rather than having an object keyed by string, you could introduce a new type, `DevicePca`, that encapsulates the devices we support:

```suggestion
type DevicePca =
  | 'PCA10028'
  | 'PCA10031'
  | ...
```

This reads as a bit more explicit to me. I believe it would also require a cast to the new type in some places, such as line 178. Possibly there is a reason you decided not to do this though?

datenreisender:
@deerob4 I had not thought about a union type. I added it in 9ac0025.

chunfantasy (approved):
https://github.com/NordicSemiconductor/nrf-device-lister-js/blob/5924e05dc672a1da90c0a27bbd99aa0cd98ff196/src/util/board-versions.js
Here are more PCAs. But can update later.

datenreisender:
> https://github.com/NordicSemiconductor/nrf-device-lister-js/blob/5924e05dc672a1da90c0a27bbd99aa0cd98ff196/src/util/board-versions.js
> Here are more PCAs. But can update later.

Yes, we can update more PCAs if they are for released products which are also used outside of Nordic. Otherwise it is ok that they just show up as “Unknown”.

PR #106, opened 2020-10-07
Update/persistent store for usage data

chunfantasy:
This PR is to add persistent store for usage data in `shared` instead of in `launcher`

### DIFF
--- a/src/persistentStore.js
+++ b/src/persistentStore.js
@@ -47,3 +47,10 @@ export const persistIsFavorite = (serialNumber, value) =>
     store.set(`${serialNumber}.fav`, value);
 export const getPersistedIsFavorite = serialNumber =>
     store.get(`${serialNumber}.fav`, false);
+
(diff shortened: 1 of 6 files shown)


src/persistentStore.js
```diff
@@ -47,3 +47,10 @@ export const persistIsFavorite = (serialNumber, value) =>
     store.set(`${serialNumber}.fav`, value);
 export const getPersistedIsFavorite = serialNumber =>
     store.get(`${serialNumber}.fav`, false);
+
+export const persistIsSendingUsageData = value =>
+    store.set('isSendingUsageData', value);
+export const getIsSendingUsageData = () =>
+    store.get('isSendingUsageData', undefined);
```
bencefr, line 54: You could use an explicit default value, false in this case, which will behave the same way as undefined, but later in the debug it looks like false instead of undefined.

bencefr (approved):

PR #111, opened 2020-10-14
Enhance lint configuration

datenreisender:
Specify the ESLint configuration in `package.json`, making it unnecessary to either copy it to the main package folder or configure other tools to pick it up at `config/eslintrc.json`.

This also makes it unnecessary to run `lint-init`, so that script is also removed from `package.json`.

I do not think there is value in releasing this as a version on its own, so I just added an entry to `Changelog.md` but not increased the version.


datenreisender:
Superseded by #113

PR #112, opened 2020-10-15
Replace moment.js with date-fns

deerob4:
As per [their own recommendation](https://momentjs.com/docs/#/-project-status/), this replaces `moment.js` with the much lighter `date-fns` library. This should remove bundle size considerably.

I'll make a corresponding PR for the launcher once this is approved and released.


datenreisender (dismissed):
Do you consider this a bug fix since you only increased the patch number of the version? We do not have strong practices around this and especially for this change it is unclear since it really does not change anything user visible (well, hopefully, only the bundle size, which is slightly visible) so I am really unsure. I just would have tended to increase the minor version but I am fine if you keep it like this.

deerob4:
@datenreisender There was a merge conflict, so I took the opportunity to bump the minor version as you suggested :)

datenreisender (approved):

PR #114, opened 2020-10-26
Add hot reloading

deerob4:
In tandem with the PR for launcher, this PR adds support for hot-loading when developing the launcher.

Here we replace the normal call to `webpack` with one to `webpack-dev-server`.

The port is fixed to 8080. It would be nice to be able to pass this in as an environment variable, but the issue then is that there is no way for the `launcher.html` to know about this (since it is just a static file, not a template). We thus can't link to the correct file in the `script` tag.

### DIFF
--- a/config/babel.config.js
+++ b/config/babel.config.js
@@ -11,7 +11,6 @@ module.exports = api => {
             '@babel/plugin-transform-spread',
             '@babel/plugin-proposal-object-rest-spread',
             '@babel/plugin-proposal-optional-chaining',
-            'react-refresh/babel'
         ],
     };
 };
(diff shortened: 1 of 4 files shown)


datenreisender:
Is the modified `webpack.config.js` here needed? As far as I understood we are only able to hot reload the launcher and the webpack configuration for that is separate in that project while the webpack configuration here is only used by the apps.

Would it be helpful for you, if we would not fix the port to 8080? I think we can make this work but I never had a need for it until now.

deerob4:
Yes, the modified webpack config is needed. If the `react-refresh` dependency is not included, then apps throw an error when launched. It's a bit counterintuitive, but this solves the problem.

Which port do you mean? As I wrote above, I can't see that there's a way to avoid fixing the `webpack-dev-server` port.

chunfantasy:
8080 is a very commonly used port. I advice not to make the port fixed to 8080.

deerob4:
Good call, I changed it to 5004

jonasem:
Might be impleneted in a different way by Pascal

PR #115, opened 2020-10-27
Autoselect a device during development

datenreisender:
Enable automatically selecting a specified device when it is detected in an app. To use this, set the environment variable AUTOSELECT_DEVICE, e.g. by running the launcher with

    AUTOSELECT_DEVICE=000680407810 npm run app

the device with the serial number 000680407810 is automatically selected when apps using the new architecture see it for the first time. 

When one deselects the device it is not automatically selected again. After restarting the app, the device is automatically selected again.

### DIFF
--- a/src/Device/deviceReducer.js
+++ b/src/Device/deviceReducer.js
@@ -88,7 +88,7 @@ const noDialogShown = {
 };
 
 const initialState = {
-    devices: [],
+    devices: {},
     selectedSerialNumber: null,
     deviceInfo: null,
     isSetupWaitingForUserInput: false,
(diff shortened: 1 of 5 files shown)


nicolai-nordic (approved):

src/Device/deviceReducer.js
```diff
@@ -88,7 +88,7 @@ const noDialogShown = {
 };
 
 const initialState = {
-    devices: [],
+    devices: {},
```
bencefr, line 91: This will break linkmonitor and tracecollector.

datenreisender: This change only affects apps using the new architecture, so I do not think it will break these apps now. 

With the new architecture the property `devices` in the state usually is an object anyhow, it was only initialised to an array which was a bug. So code for the new architecture always has to handle it being an object. 

But apps must never read the part of the redux state directly anyhow that is handled by `shared`. The shape of the redux state is not part of the API. If apps need to know about the shared state, we will have to provide selectors for it that the apps then can use.

bencefr: You are right, it will not break these apps as the reducer comes from the legacy code. However there it is a List() which I assume was reimplemented as an array, not an object.
I'm also not arguing what an app should or shouldn't do... but you see, sometimes "it's more of a guideline than an actual rule". :)

bencefr (approved):

PR #119, opened 2020-11-05
Exclude node-modules from eslint-loader to prevent warning bonanza

nicolai-nordic:
Build logs get spammed with warnings because eslint-loader is looking through node_modules and failing to locate files. So split up babel-loader and eslint-loader rules so that we can exclude node_modules correctly.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -80,6 +80,13 @@ module.exports = {
                                 './node_modules/pc-nrfconnect-shared/config/babel.config.js',
                         },
                     },
+                ],
+                exclude: /node_modules\/(?!pc-nrfconnect-shared\/)/,
+            },
+            {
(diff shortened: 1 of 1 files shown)


datenreisender (approved):
Looking at the change I am still puzzled why it is needed. 

My understanding of the old rule is, that it already excludes most of `node_modules` just including `node_modules/pc-nrfconnect-shared` (which is needed because we do not precompile the files in `shared`, so the projects that use `shared` need to compile them). 

I had shortly assumed, that the flaw of this rule is that files in `node_modules/pc-nrfconnect-shared/node_modules` may cause the issue but at least the error states that the problem is in `node_modules/@popperjs`. So that assumption also seems wrong. Strange.

But since this change seems to fix it, let us use that.

PR #126, opened 2020-11-25
Enhance components

datenreisender:
Lots of small changes, also [documented in the `Changelog.md`](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/e3b001be43631a452ce4ba18356b5b73ab5cb4b5/Changelog.md#4160):

### Disabling `Slider` and `InlineInputs` as well as styling other elements as disabled
This code:

```js
<Form.Label className="disabled" htmlFor="length">
    <span className="flex-fill">Length</span>
    <NumberInlineInput disabled value="10" range={range} /> ms
</Form.Label>
<Slider
    disabled
    id="length"
    values={[10]}
    onChange={[() => {}]}
    range={range}
/>
```
now works as expected (notice the three instances of `disabled`). Here are the old and new looks:

ScreenCapture at Wed Nov 25 14:52:47 CET 2020 ScreenCapture at Wed Nov 25 14:53:23 CET 2020

### Corrected styling of disabled `Toggle`
The styling of disabled `Toggle`s was changed. This is the before and after:

Screenshot 2020-11-25 at 14 57 17 Screenshot 2020-11-25 at 14 57 41

### Convenience function `classNames`

A convenience function `classNames` was added to construct compound class names. It filters out all values that are not strings. The idea of this function is to use it with conditionals and potentially unset values like this:
```js
classNames(
     'fixed-class-name',
     isVisible && 'visible',
     isEnabled ? 'enabled' : 'disabled',
     potentiallyUndefined,
)
```

### Property `active` on all rendered panes
A property `active` is set on all rendered panes that is only for the currently active true and false for all others

### Ticks on a `Slider`
This code:
```js
<Slider
    ticks
    values={[3]}
    range={{ min: 0, max: 5 }}
    onChange={[() => {}]}
/>
```
renders to this:

Screenshot 2020-11-25 at 15 01 47

### Dynamic `InlineInput` length

The components `InlineInput` and `NumberInlineInput` grow and shrink depending on their current content. This made the property `chars` unnecessary and it is now ignored.


nicolai-nordic (approved):
Great job!

PR #127, opened 2020-11-26
Fix: Stacked opacity on nested disabled elements

datenreisender:
The opacity of disabled elements stacked up when they were nested. E.g.
in the following code the input had the opacity applied twice (making the
opacity squared), so it looked lighter than than supposed:
```jsx
       <div className="disabled">This <InlineInput disabled value="doubled"/></div>
```


chunfantasy (approved):
LGTM

PR #130, opened 2021-01-25
Small fixes

datenreisender:
- At least on macOS (unsure about Linux, is no problem on Win 10) clicking on URLs in the log viewer did not open it in a browser
- In the PPK app we need to query and set the active pane and I think this also already came up in other apps. But this was not possible through an official API yet, so this adds an API for that.
- The `systeminformation` package had a security vulnerability for which @dependabot created #116. This makes that PR obsolete.

### DIFF
--- a/src/Log/LogEntry.jsx
+++ b/src/Log/LogEntry.jsx
@@ -36,8 +36,9 @@
 
 import React from 'react';
 import formatDate from 'date-fns/format';
-import { shell } from 'electron';
 import { number, shape, string } from 'prop-types';
+
+import { openUrl } from '../utils/open';
 
 import './log-entry.scss';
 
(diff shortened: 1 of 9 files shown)


bencefr (approved):
LGTM

src/Log/LogEntry.jsx
```diff
@@ -66,7 +67,7 @@ function hrefReplacer(str) {
                 href={href}
                 key={index}
                 tabIndex={index}
-                onClick={() => shell.openItem(href)}
+                onClick={() => openUrl(href)}
```
bencefr, line 70: I see this is inherited from old architecture. This will call `shell.openExternal()`, does that work properly for all platforms?

datenreisender: We already use the same function in other places: Clicking on the logo or on the about pane. It better works on all platforms, otherwise it is broken in those places too. ;-) But I double checked on Windows.

PR #131, opened 2021-01-26
Add nrf-device-lib-js as extern

deerob4:


datenreisender:
The PR is named “Add nrf-device-lib-js as extern” but the code looks like more than this is happening here. Could you describe that briefly? 

Also: Would it be helpful to also describe in the Changelog what these changes in this PR mean for the launcher or apps? Or is this just too much work in progress and an intermediate step right now, that it is mostly meaningless to describe it any further?

chunfantasy:
Yep, I continued on a very early PR so the some of the descriptions are out of date. I will update the info you mention above.

datenreisender (approved):

PR #139, opened 2021-03-12
Gather common functionality

datenreisender:
This is part of [NCP-3569](https://projecttools.nordicsemi.no/jira/browse/NCP-3569).

Sorry for the big PR that collects too many different things. I am afraid this also makes reviewing harder for you. :-/ 

The changes visible to the outside are described in the Changelog:
> ### Added
> - Shared styles can now be imported in SCSS from
>   `~pc-nrfconnect-shared/styles`.
> - When users start an app, the pane is opened again, that was open
>   when they left the app the last time. Only the about pane is ignored,
>   because we assume people are not interested in returning to the about
>   pane.
> ### Changed
> - Types for the exported colors got more specific.
> - The format for specifying the panes of the app as a property to the
>   `App` component has changed. See below for details.
> ### Steps to upgrade when using this package
> - If you had an import like
>   ```scss
>   @import '~pc-nrfconnect-shared/src/variables';
>   ```
>   in your SCSS code before, you should replace it with
>   ```scss
>   @import "~pc-nrfconnect-shared/styles";
>   ```
>   because even though for now the code with
>   `~pc-nrfconnect-shared/src/variables` will keep on working, it is
>   not part of the public API that we try to preserve. Contrary to
>   `~pc-nrfconnect-shared/styles`, which is part of the supported API.
> - Previously panes were defined as a pair of pane name and component,
>   like this:
>   ```js
>   panes={[
>     ['Dashboard', Dashboard],
>     ['Terminal', Terminal],
>   ]}
>   ```
>   This was changed to an array of objects, like this:
>   ```js
>   panes={[
>     { name: 'Dashboard', Main: Dashboard, },
>     { name: 'Terminal', Main: Terminal, },
>   ]}
>   ```
>   The old format still is supported but will issue a warning and will
>   be removed in the future.

Other, internal changes:
- d22a1a0 Replaced DYI implementation for resize detection
- 6895845 Fixed: Type for device icon was wrong
- 213c4c1 Corrected dependencies
- 8bd160e Explain a disabled lint warning
- 178b87f Unified naming in Changelog


nicolai-nordic (approved):
Looks great! Tested persisting panes and that works as intended. Also it correctly issues a warning about legacy pane usage. Great job!

I also liked that you added babel nullish coalescing plugin.

I haven't been able to verify that `useResizeDetector` correctly replaces `useWidthObserver`, since I don't have any hardware available.

PR #142, opened 2021-03-25
Use the app's package.json

datenreisender:
This is part of [NCP-3569](https://projecttools.nordicsemi.no/jira/browse/NCP-3569).

https://github.com/NordicSemiconductor/pc-nrfconnect-shared/pull/139/commits/089dc10d3bf83eba2da4b96136f817b75f0b6d59 in #139 previously added the ability to `shared` to read and use the content of the app's `package.json`. This PR uses this for two things: usage data and the persistent store, as described in the changelog:

### Added
- Property `reportUsageData` to component App.
- Function `getPersistentStore` to get a persistent store, specific for the
  app. The app name from `package.json` is used to identify the app.
### Steps to upgrade when using this package
- If your app wants to report usage data, you can remove the code to call
  `usageData.init()` and instead set the property `reportUsageData` to the
  component `App`.
- If your app already use a persistent store, you can switch to
  `getPersistentStore()`, just be aware that if you currently use a different
  name for the store than the app name from `package.json`, then the old
  settings are either lost of you have to care for migrating them.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,16 @@
 ## 4.21.0
 ### Added
 - Property `reportUsageData` to component App.
+- Function `getPersistentStore` to get a persistent store, specific for the
+  app. The app name from `package.json` is used to identify the app.
 ### Steps to upgrade when using this package
 - If your app wants to report usage data, you can remove the code to call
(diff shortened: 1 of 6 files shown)


nicolai-nordic (approved):
LGTM!

PR #144, opened 2021-04-09
Feature/card component

helenelonvik:
In this PR a card component is implemented. 
It can be used where current or new cards are used in the apps (f.eks in about pane)


datenreisender (approved):
I made some changes:

- a436eba: Just some minor cleanups, look in the commit for details.
- d623434: All our JavaScript files have a copyright header, so this adds that.
- a8670af: Changes in shared should be described in `Changelog.md`, so that people upgrading to a new version of shared know what they get. And in order to reference a new version of shared we should also bump the version number. (After this branch is merged we also need to tag it and create a release on GitHub. We can do that together as soon as you have merged this)
- 324796d: Renamed the `CardComponent` to just `Card`, because we usually do not name things `…Component`. Otherwise we should name all our components `…Component` and that would be just noise.
- b2e3291: While it is possible to have two components just called `Card`, it could create some confusion. So I renamed the more specific component to `AboutCard`, since it is just for cards on the About pane.

Please check that my changes did not break anything.

As I commented in `src/Card/card.scss` I suggest that you do another change there. But for me it is fine if you first merge this PR as it is and work on that change afterwards.

This PR is blocked by the CLAassistant because you used two different email addresses in commits: helenetl@stud.ntnu.no and helene_t_l@hotmail.com (previously you also used helene.lonvik@nordicsemi.no, btw). I assume that happened because you had to switch to a different computer last week. You can either (as the comment https://github.com/NordicSemiconductor/pc-nrfconnect-shared/pull/144#issuecomment-820409757 by CLAassistent suggests) add the missing email address to your GitHub account or we can change the author in the commits. If you want to choose the latter and are uncertain how to do this, I can assist you with it or do that for you.

You use these new cards now only in the Dashboard of the CM app. But this means, that we would have pretty different cards in the Dashboard and in the About pane. I think it would be better to have a more uniform look.

I suggest that you next work on using the Card component also in the About pane. But that, of course, leads to the question in which way we want to change the styling of the About pane (or the Dashboard) in such a way, that the two look similar. Please discuss with Ketil, what kind of look he would like to have in these two. Maybe, to have first something more concrete to discuss with him, change the current About pane so that it uses your new Card component (which might make the current `AboutCard` obsolete) and show to Ketil how that would look and whether he likes that. If not, then decide with him on a look that works in both places.

So, since this was quite long, this is what I think you should do next:

- [x] Check whether my changes did not break anything
- [ ] Fix the block by the CLAassistant
- [ ] If you like to, merge this PR in it’s current state and create a release.
- [ ] Use the new Card component in the About pane in a way that is agreed on with Ketil.
- [ ] Fix that currently in `src/Card/card.scss` it is styled how the card is embedded.

datenreisender (approved):

PR #145, opened 2021-04-15
Prevent esModule syntax from url-loader

nicolai-nordic:
In a recent major release of `url-loader` it started using ES module syntax by default. This breaks some of our imports, so we prefer to not utilize this behavior for now.


datenreisender (approved):
As discussed, this may not be necessary if fixed in all apps, but we could still also apply this fix here. Your choice, whether you want to merge it.

PR #147, opened 2021-04-28
Explain automated build step

datenreisender:


datenreisender:
Closing PR because the build fails and I rather include the change from this PR into another PR which makes the build succeed again and also fixes the problem behind that.

PR #149, opened 2021-05-03
Fix: Type of Toggle props were slightly wrong

datenreisender:
- `onToggle` is optional. It can be `undefined`, to use the toggle
  just for decoration or to show a current value.
- `barToggledColor` was wrong, correct is `barColorToggled`.
- `label` is optional. When `children` are specified the `label` is not
  used anyhow. If neither `children` nor a `label` is specified, then
  there simply will be no label.


chunfantasy (approved):
Good good!

PR #156, opened 2021-05-28
Add new FactoryResetButton component.

nicolai-nordic:
This implements [https://trello.com/c/li9bqFPm/49-factory-reset-in-the-error-message-about-pane-other-places](https://trello.com/c/li9bqFPm/49-factory-reset-in-the-error-message-about-pane-other-places).

This PR adds a `FactoryResetButton` component. This component provides a confirmation dialog to confirm that a reset should be done. Also users of this component can override the default reset functionality and the confirmation message. This button is also added to the About pane, so users have a handy way of restoring default settings.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+## 4.24.0
+### Added
+- Component `FactoryResetButton`. This component is also added to the `About` pane.
+
 ## 4.23.0
 ### Added
 - Error robustness
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "4.23.1",
+    "version": "4.24.0",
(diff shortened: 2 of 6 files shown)


src/About/ApplicationCard.jsx
```diff
@@ -71,6 +72,12 @@ export default () => {
                     label="Get source code"
                 />
             </Section>
+            <Section>
+                <FactoryResetButton
+                    label="Restore defaults"
+                    modalText="foo bar"
```
chunfantasy, line 78: Displaying foo bar looks a bit weird.

nicolai-nordic: Oh oops! Good thing you caught that!

chunfantasy (approved):
Good job!

chunfantasy (approved):

PR #158, opened 2021-06-08
Update dependencies

datenreisender:
Update [`systeminformation`](https://github.com/sebhildebrandt/systeminformation) (for which there is [a security alert](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/security/dependabot/package.json/systeminformation) even though it probably does not affect us) and several webpack related packages. Including migrating from the deprecated [`eslint-loader`](https://github.com/webpack-contrib/eslint-loader) to [`eslint-webpack-plugin`](https://github.com/webpack-contrib/eslint-webpack-plugin).

Also fix a bug in generating a system report from the about pane.


chunfantasy (approved):
LGTM

PR #159, opened 2021-06-09
Feature/card component

helenelonvik:
This PR will update about cards in the about pane to use the card component i shared. 
It also update the component some


nicolai-nordic (approved):
LGTM!

PR #172, opened 2021-08-19
feat: add drowdown component

PascalPickel:
Signed-off-by: Pascal Pickel <Pascal.Pickel@nordicsemi.no>


datenreisender:
Please add a changlog entry describing this change. 

Also: Did you push a test branch of an app that utilises this change so that I can easily see that/how it works there?

PascalPickel:
> Please add a changlog entry describing this change.
> 
> Also: Did you push a test branch of an app that utilises this change so that I can easily see that/how it works there?

https://github.com/NordicSemiconductor/pc-nrfconnect-dtm/tree/test/Dropdown

index.d.ts
```diff
@@ -428,6 +428,18 @@ declare module 'pc-nrfconnect-shared' {
 
     export class Toggle extends React.Component<ToggleProps> {}
 
+    // Drowdown.jsx
+
+    interface DropdownProps {
+        label: string;
+        items: string[];
+        onSelect: (item: number) => void;
```
jonasem, line 436: Wouldn't it be handier on most cases that this returned the item?

PascalPickel: > Wouldn't it be handier on most cases that this returned the item?

Was considering that at first but figured that using an index would be simpler.
When you have an array or dictionary like structure of strings that are only used for display purposes (https://github.com/NordicSemiconductor/pc-nrfconnect-dtm/blob/master/src/SidePanel/PacketView.tsx#L79), it would require the developer to write some extra .find() code.

Maybe I'm just over complicating it though so let me know if you still think the item return is better.

datenreisender: Hard to predict. Sometimes the item is handy, sometimes the index. The latter may happen especially if the item is just a label representing another object and if it is not so easy to infer from the label back to that object. 

At least [in Pascals exemplary use of the component](https://github.com/NordicSemiconductor/pc-nrfconnect-dtm/compare/test/Dropdown#diff-51b51e853384fb68648bfc876ab0a90a36ead8ebe3f0264fb2db8a688d1f4bb0) the index works pretty well. The [use of a Dropdown in the CM app](https://github.com/NordicPlayground/pc-nrfconnect-cellularmonitor/blob/77d616bf0741d76056b72cc24a5dadea0ed77b67/src/components/SidePanel/Serialports.tsx) looks like there the item might be more handy.

Maybe we can take the route of the built-in iteration functions (like `forEach` and `map`) and return both: The item and the index.

jonasem: I too failed to realize that the items passed in are as a string array, and not as the objects the list is refering to. I'm used to from some other libraries that the objects of the dropdown can be anything, and that a selector for a label is passed as well. So then I agree it's more useful sending the index back as in this case now :)

PR #173, opened 2021-08-19
Update/inlineinput style

PascalPickel:


datenreisender:
Does this change the visuals in any way or keep it completely the same?

PascalPickel:
> Does this change the visuals in any way or keep it completely the same?

Before (size + 2)
Screenshot from 2021-08-20 11-49-07

After (no additional spacing and min 1)
Screenshot from 2021-08-20 11-50-03

nicolai-nordic (approved):

PR #182, opened 2021-08-31
Fix/css issue after chromium upgrade

nicolai-nordic:
This PR implements a fix for https://trello.com/c/IIuVvkIP/232-bug-possibly-related-to-log-in-apps-using-the-new-architecture

Likely this was caused by [this issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1150451#c9) introduced after upgrading Chromium past v86. How the bug looks in our apps can be seen [here](https://www.loom.com/share/3e531fe4dd2f4e7a8cfc3be3112d8c06).

I am not sure about the best approach to fix issue, but for this fix I tried using `max-height` instead of `margin-bottom` to control the size/position of the log window, which looks good to me at least.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "4.29.0",
+    "version": "4.29.1",
     "description": "Shared commodities for developing pc-nrfconnect-* packages",
     "repository": {
         "type": "git",
--- a/src/App/app.scss
+++ b/src/App/app.scss
@@ -1,4 +1,4 @@
-@import "../variables";
+@import '../variables';
 
 .core19-app {
     background-color: $gray-50;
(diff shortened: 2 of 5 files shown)


chunfantasy (approved):
LGTM

src/App/app.scss
```diff
@@ -42,12 +42,16 @@
             overflow: hidden;
             width: 100%;
 
-            margin-bottom: 0;
-            transition: margin-bottom $normal-transition;
+            max-height: $log-viewer-height;
```
datenreisender, line 45: I assume changing the animated property from `margin-bottom` to `max-height` does change the form of animation from sliding out the log to shrinking it. Is that correct and are you fine with that?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "4.29.0",
+    "version": "4.29.1",
```
datenreisender, line 3: It seems like 4.29.0 was not yet released, so there is no need to bump the version.

src/App/app.scss
nicolai-nordic: It does, but the difference wasn't really perceptible as far as I could tell. But that could just be me, so we can look at improving that later, if this change is in fact noticeable.

PR #183, opened 2021-09-08
fix: display thingy91 device name in select device

aadnekar:
The thingy91 does not display a boardVersion, and therefore it is not
properly mapped to the PCA. This commit implements a soltuion to detect
the thingy91 from the serialNumber instead.

## Previously Unknown
image

## Fix now display correct device name
image


nicolai-nordic (approved):
Looks good!

datenreisender (approved):

PR #185, opened 2021-09-09
Mock and lint

nicolai-nordic:
Multiple minor changes:

- 5e61c26: Remove unused modules from eslint config
- abfaaa0: Mock `nrf-device-lib-js` and add `moduleNameMapper` entry in jest config
- 2f08c08: `nrf-device-lib-js` types changed slightly, so update accordingly.
- 04d70d6: For the BLE app, the firmware validation is handled slightly different, so update the `validateFirmware` function to handle this case as well.

### DIFF
--- a/src/Device/jprogOperations.js
+++ b/src/Device/jprogOperations.js
@@ -151,11 +151,12 @@ async function validateFirmware(device, fwVersion) {
             typeof fwVersion.validator === 'function'
         ) {
             valid = fwVersion.validator(fwInfo.imageInfoList);
+        } else {
+            valid = fwInfo.imageInfoList.find(imageInfo => {
+                if (!imageInfo.version.string) return false;
+                return imageInfo.version.string.includes(fwVersion);
+            });
         }
-        valid = fwInfo.imageInfoList.find(imageInfo => {
-            if (!imageInfo.version.string) return false;
-            return imageInfo.version.string.includes(fwVersion);
-        });
         await reset(device.id); // Remove after fixing in nrf-device-lib
         return valid;
     } catch (error) {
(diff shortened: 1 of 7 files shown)


src/Device/jprogOperations.js
```diff
@@ -138,20 +140,27 @@ const verifySerialPortAvailable = device => {
...
             device.id
         );
-        const valid = fwInfo.imageInfoList.find(imageInfo =>
-            imageInfo.version.string.includes(fwVersion)
-        );
+        if (
+            typeof fwVersion === 'object' &&
+            typeof fwVersion.validator === 'function'
+        ) {
+            valid = fwVersion.validator(fwInfo.imageInfoList);
```
jonasem, line 153: This one is overwritten by the next line after the if block.

src/Device/jprogOperations.js
```diff
@@ -138,20 +140,27 @@ const verifySerialPortAvailable = device => {
...
+        }
+        valid = fwInfo.imageInfoList.find(imageInfo => {
+            if (!imageInfo.version.string) return false;
+            return imageInfo.version.string.includes(fwVersion);
+        });
         await reset(device.id); // Remove after fixing in nrf-device-lib
-        if (valid) return true;
+        return valid;
```
jonasem, line 160: Could this simply be
```
return fwInfo.imageInfoList.find(imageInfo => {
            if (!imageInfo.version.string) return false;
            return imageInfo.version.string.includes(fwVersion);
        });
```

jonasem (approved):

src/Device/jprogOperations.js
nicolai-nordic: Oops, good catch, there was supposed to be an else there, wonder if auto-formatting changed that somehow.

src/Device/jprogOperations.js
nicolai-nordic: Originally that would be the case, but then we discovered that the BLE app validates the firmware differently, i.e. `fwVersion` is an object with a `validator` function, not a string, so this comparison would always fail in the BLE app.

PR #188, opened 2021-09-13
Turn off rule to allow explicit export without default

jonasem:
Turn of rule that requires a default export if there is only a single export. No code is expected to be able to break from this change.


nicolai-nordic (approved):
Yay!

PR #194, opened 2021-09-22
Typescript reducers to slices

jonasem:


index.d.ts
```diff
@@ -9,39 +9,16 @@ declare module 'pc-nrfconnect-shared' {
     import React from 'react';
     import winston from 'winston';
     import Store from 'electron-store';
-    import { Device as NrfdlDevice } from '@nordicsemiconductor/nrf-device-lib-js';
+
+    type RootState = import('./src/state').RootState;
+    type Device = import('./src/state').Device;
+    type DeviceInfo = import('./src/state').DeviceInfo;
+    type Serialport = import('./src/state').Serialport;
```
datenreisender, line 16: Woah, using dynamic imports here is an interesting idea and yet it surprised me that this works. How did you come up with this idea? Is it documented somewhere that this is supposed to work this way?

I am surprised, because [the documentation of dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports) says that they are supposed to return a Promise, not a direct value. But of course that documentation talks about JavaScript files not TypeScript declaration files, so I have no idea whether this works here just by chance or by design. 

Also, as you probably know, normally ambient module declarations annoyingly cannot reference modules through relative module name, e.g. `import type { Serialport } from './src/state';` is forbidden here. So this makes me even more curious why this works.

datenreisender (changes requested):
I am afraid this does not work, because the initial states are computed too early. See my comment on that in `src/App/appLayout.ts`.

PR #197, opened 2021-09-27
Mixed bag

datenreisender:
A mixed bag of mostly small changes:
- 5fb4159: The DeviceSelector already supported the property `allowCustomDevice` in the `DeviceSetup`, it was just missing in the type and prop types.
- 42a0317: When using custom devices (which can be done in the BLE app), the setup was not ran correctly.
- 5c84ddc: Small change which allows passing children to the component `App`. This is useful for components outside the normal render flow (e.g. dialogs) or invisible components, which are rendered just for other means (e.g. to dispatch Redux actions when the App is started), as explained in `Changelog.md`.
- 5acb2ab: Enables two additional ESLint rules: [`require-await`](https://eslint.org/docs/rules/require-await) and [`react/jsx-key`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md).

### DIFF
--- a/src/Device/deviceSetup.js
+++ b/src/Device/deviceSetup.js
@@ -153,6 +153,17 @@ export const prepareDevice = async (device, deviceSetupConfig) => {
     return createReturnValue(device, { wasProgrammed }, detailedOutput);
 };
 
+const onSuccessfulDeviceSetup = (
+    dispatch,
+    device,
+    doStartWatchingDevices,
+    onDeviceIsReady
+) => {
+    doStartWatchingDevices();
+    dispatch(deviceSetupComplete(device));
+    onDeviceIsReady(device);
+};
+
 /**
  * Selects a device and sets it up for use according to the `config.deviceSetup`
  * configuration given by the app.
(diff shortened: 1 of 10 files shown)


nicolai-nordic:
I actually needed  to set `allowCustomDevice` in the DTM app today, so good timing with this fix!

src/Device/deviceSetup.js
```diff
@@ -196,12 +207,35 @@ export const setupDevice =
...
+                error.message.includes('No firmware defined')
+            ) {
+                logger.info(
+                    `Connected to device with serial number: ${device.serialNumber} ` +
+                        `and family: ${device.deviceInfo.family || 'Unknown'} `
+                );
+                logger.debug(
+                    'Note: no pre-compiled firmware is available for the selected device. ' +
+                        'You may still use the app if you have programmed the device ' +
+                        'with a compatible connectivity firmware.'
```
nicolai-nordic, line 229: I'd change it from `compatible connectivity firmware` to just `compatible firmware` since this message can be shown in multiple apps (e.g. DTM)

nicolai-nordic (approved):
Nice changes!

src/Device/deviceSetup.js
datenreisender: Good idea. I mindlessly [copied it from the BLE app](https://github.com/NordicSemiconductor/pc-nrfconnect-ble/blob/57388a4451a3c21a2b48462b07263d37e2968520/index.jsx#L129-L133) but your wording is better. Will change that.

datenreisender: I will also change the log level from `debug` to `info`, because I consider this message to be helpful for end users, especially struggling to understand why an app that supports custom devices does not work with their device.

jonasem (approved):

PR #200, opened 2021-10-01
feat: add documentation card in about pane

PascalPickel:
Adds a documentation property to the App component which will show "documentation sections" in the about pane.

Screenshot from 2021-10-01 14-15-52

An array of objects like these would be passed to the App component:
```
<DocumentationSection
                title="Infocenter"
                link="https://infocenter.nordicsemi.com/index.jsp?topic=%2Fug_nc_programmer%2FUG%2Fnrf_connect_programmer%2Fncp_introduction.html"
                linkLabel="Go to infocenter"
            >
                For more information about this app and its usage, visit the
                infocenter page for it.
</DocumentationSection>
```

### DIFF
--- a/src/About/documentationSlice.js
+++ b/src/About/documentationSlice.js
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-4-Clause
+ */
+
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {
+    sections: [],
+};
+
+const slice = createSlice({
+    name: 'documentation',
+    initialState,
+    reducers: {
+        setDocumentationSections: (state, action) => {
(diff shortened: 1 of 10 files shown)


src/About/documentationSlice.js
```diff
@@ -0,0 +1,29 @@
+/*
```
jonasem, line 1: This has no dependencies, maybe rather make it directly in typescript now, instead of having to rewrite later.

src/About/documentationSlice.ts
```diff
@@ -0,0 +1,40 @@
...
+    title: string;
+    linkLabel: string;
+    link: string;
+}
+
+interface DocumentationState {
+    sections: DocumentationSection[];
+}
+
+const initialState = {
```
jonasem, line 19:
```suggestion
const initialState: DocumentationState = {
```

jonasem:
Add note in changelog and bump version

jonasem (approved):

PR #205, opened 2021-10-11
Update/module def

nicolai-nordic:
- Update module definitions
- Change to CSS modules
- Add some tests

(Also fix one bug with very long error messages breaking Error Boundary layout)


chunfantasy (approved):
LGTM

PR #218, opened 2021-10-22
Rewrite remaining types

jonasem:
This is not compatible with legacy launcher.


src/utils/logLibVersions.ts
```diff
@@ -73,8 +73,9 @@ export default async () => {
         logVersion(versions, 'nrfjprog_dll', 'nrfjprog dll');
         logVersion(versions, 'jlink_dll', 'JLink');
     } catch (error) {
-        logger.error(
-            `Failed to get the library versions: ${error.message || error}`
-        );
+        const message = error instanceof Error ? error.message : error;
+        if (error instanceof Error) {
+            logger.error(`Failed to get the library versions: ${message}`);
+        }
```
nicolai-nordic, line 79: Why do we only log if it's an instance of `Error`? We still have the message right?

src/Device/jprogOperations.ts
```diff
@@ -109,11 +114,11 @@ const verifySerialPortAvailable = device => {
  * @param {String} fwVersion The firmware version to be matched.
  * @returns {Promise} Promise that resolves if successful or rejects with error.
  */
-async function validateFirmware(device, fwVersion) {
-    let valid = false;
-    let fwInfo;
+export async function validateFirmware(device: Device, fwVersion: any) {
```
nicolai-nordic, line 117: TS type says `any` but jsdoc says `String`?

nicolai-nordic (approved):

src/utils/logLibVersions.ts
jonasem: Typescript won't let us access the message property before checking the type of the error. In theory "anything" can be the type of the exception caught.

src/Device/jprogOperations.ts
jonasem: Typescript doesn't like that we do a `typeof fwVersion.validator === 'function'`. String doesn't have a validator property, so jsdoc must be wrong.

src/Device/sdfuOperations.ts
```diff
@@ -125,11 +138,14 @@ const confirmHelper = async promiseConfirm => {
...
-const choiceHelper = /* async */ (choices, promiseChoice) => {
+const choiceHelper = /* async */ (
+    choices: string[],
+    promiseChoice: PromiseChoice
+) => {
     if (choices.length > 1 && promiseChoice) {
         return promiseChoice('Which firmware do you want to program?', choices);
     }
-    return choices.pop();
+    return choices.slice(-1)[0];
```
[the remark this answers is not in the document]
jonasem: It seems strange to mutate the choices when it is used later by the calling method.

src/Device/jprogOperations.ts
nicolai-nordic: Oh right, this is dependent on the app. I think for BLE it's an object with a validator `property` plus maybe some other things, while for other apps it's a string. So it should probably be of type `{ validator: Function, ... } | string`

src/Device/deviceLister.ts
```diff
@@ -143,28 +158,23 @@ export const stopWatchingDevices = () => {
...
     timeout = DEFAULT_DEVICE_WAIT_TIME,
-    expectedTraits: DeviceListing = { serialPort: true }
+    expectedTraits: DeviceTraits = DEFAULT_TRAITS
 ) => {
     logger.debug(`Will wait for device ${serialNumber}`);
 
     // To be compatible with `serialport`
-    expectedTraits = {
-        serialPort: expectedTraits.serialport,
-    };
```
datenreisender, line 159: I am uncertain about this change, because it really looks like the new code does something else than the old code and my understanding of the PR was, that you did not want to change behaviour. 

But while the old code replaced what was in `expectedTraits` with a new object with just the property `serialPort` in it, the new code adds the property `serialPort` to the object, but also leaves all existing properties there. 

Is this correct?

PR #232, opened 2021-11-12
Port react jsx files to typescript

jonasem:
Rewrite the remaining files (jsx) to typescript. After this PR all the code, except tests will be typed.


src/About/DeviceCard.tsx
```diff
@@ -49,12 +49,13 @@ export default () => {
             <Section title="ID">{device.serialNumber}</Section>
             <Section title="PCA">{pca || 'Unknown'}</Section>
             <Section title="Cores">{cores || 'Unknown'}</Section>
+            {/*
```
jonasem, line 52: This is an existing bug, doesnt work today. Commented out and made a trello https://trello.com/c/9XBs3jZF/169-fix-device-info-in-about-pane

src/Device/DeviceSetup/DeviceSetupView.tsx
```diff
@@ -37,13 +49,17 @@ export default class DeviceSetupDialog extends React.Component {
             this.props;
         const { selectedChoice } = this.state;
 
-        if (choices && (choices.length > 0 || choices.size > 0)) {
+        if (
+            choices &&
+            (choices.length > 0 ||
+                (choices as unknown as { size: number }).size > 0)
+        ) {
```
nicolai-nordic, line 56: Can `choices` be either a list of strings or an object with a `size` property?

nicolai-nordic: Is this some remnants from `Immutable.js` or something that we can just remove?

nicolai-nordic (approved):
Very long PR, so it's difficult to get a good overview, but it looks good from what I can see.

src/Device/DeviceSetup/DeviceSetupView.tsx
jonasem: Good catch! It's actually not exported, so no other apps are using it directly, and internally in shared we are only using string.

datenreisender:
I am too late to review this. But the next time, please do not leave a bunch of those WIP commits behind. More expressive commit descriptions would, of course, be helpful. Maybe squash them, if they do not make sense on their own.

PR #235, opened 2021-11-18
Enhance error logging

datenreisender:
In #218 in lots of error logging places the code `error.message` was replaced by `error instanceof Error ? error.message : error`. This altered the behaviour because e.g. device-lib returns error objects which do have a `message` property but are not instances of `Error`.

#233 enhanced the situation by calling `JSON.stringify` in one place but it also means that the whole error object is now dumped as JSON into the log, which may not be so readable.

This PR now wants to mostly restore the initial behaviour and enhance it slightly:
1. A function `logError` on the logger is introduced, so that everywhere where we want to log an error (usually in a `catch` block) we can use the same function.
2. That function handles different kind of situations: 
    1. If the error is an instance of `Error`, then the error itself is printed, because JavaScript Error objects usually provide a good `toString` implementation.
    2. If the error has a `message` property, then that is printed.
    3. If the above do not apply, then use `JSON.stringify` on the error object.
    4. If the error contains a property `origin` (like some errors from device-lib) then print that also.
3. There is [a unit test for `logError`](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/f89323d9b18bfa7d313268dcaec6d9a99a317b16/src/logging/createErrorMessage.test.ts) to illustrate and test the different behaviours.
4. In shared the function `logError` is now used in all places where it seems reasonable.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Fixed
 - [Object object] error message when device enumeration failed.
+### Added
+- Function `logError` to ease logging error.
 
 ## 5.8.0 - 2021-11-08
 ### Added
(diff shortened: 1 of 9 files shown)


nicolai-nordic (approved):
Really nice! +1 for tests also.

PR #236, opened 2021-11-18
Return device even when not programming it

jonasem:
It seems that apps expect the shared device state to have a device in the deviceInfo property.

### DIFF
--- a/src/Device/deviceSetup.ts
+++ b/src/Device/deviceSetup.ts
@@ -108,7 +108,7 @@ export interface DeviceSetupConfig extends DeviceSetup {
 export const prepareDevice = async (
     device: Device,
     deviceSetupConfig: DeviceSetupConfig
-): Promise<Device | undefined> => {
+): Promise<Device> => {
     const { jprog, dfu, needSerialport } = deviceSetupConfig;
 
     if (dfu && Object.keys(dfu).length > 0) {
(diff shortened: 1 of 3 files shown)


src/Device/deviceSetup.ts
```diff
@@ -228,13 +224,10 @@ export const setupDevice =
...
             );
-
-            if (preparedDevice === undefined)
-                throw new Error('Unable to prepare device');
-
+            
             onSuccessfulDeviceSetup(
                 dispatch,
-                preparedDevice,
+                preparedDevice ?? device,
```
nicolai-nordic, line 230:
```suggestion
                preparedDevice,
```

src/Device/jprogOperations.ts
```diff
@@ -183,17 +183,17 @@ export async function programFirmware(
...
-        await program(device.id, fw);
-        logger.debug(`Resetting ${device.serialNumber}`);
-        await reset(device.id);
+        if (confirmed) {
+            logger.debug(`Programming ${device.serialNumber} with ${fw}`);
+            await program(device.id, fw);
+            logger.debug(`Resetting ${device.serialNumber}`);
+            await reset(device.id);
+        }
+        return device;
```
nicolai-nordic, line 192: Revert/update this to early return `device` if the user declines and move the return of the device to after the `try/catch`.

nicolai-nordic (approved):

PR #240, opened 2021-12-07
Expose test tools

datenreisender:
Main intent of this: Expose some tools as an object `testUtils` to ease testing.

In details these are:

- `rootReducer`: Takes the app reducer as an argument and then creates the root reducer, like it is used in the running app. This is often needed when testing reducers or React components in apps, like with the following two functions.
- `dispatchTo`: Dispatch a sequence of actions to a reducer and get the resulting state. An example where this is used, is in [the currently developed `traceSlice.test.ts` in CM](https://github.com/NordicPlayground/pc-nrfconnect-cellularmonitor/blob/some_enhancements/src/features/tracing/traceSlice.test.ts#L107-L111). 

  I am not sure, whether the type will also work for thunks, maybe we have to cast those first or refine the type of `dispatchTo` later, but I was not able to achieve that now.
- `render`: Render a component with [React Testing Library](https://testing-library.com/react), attached with a Redux provider, using a specified reducer and after dispatching a specified sequence of actions. 
 
  Because the reducer will often be the same for the whole app, this function is called once just with a reducer and the returned function can then be called with the React element (and optionally an array of actions). E.g. [here it is called just with a reducer](https://github.com/NordicPlayground/pc-nrfconnect-cellularmonitor/blob/980ad15f70f4246bc2b749429d8b968c84aa7a74/src/utils/testUtils.tsx#L38) and [the returned function is then used to test a React element after dispatching `setTraceIsStarted` and `setTraceProgress`](https://github.com/NordicPlayground/pc-nrfconnect-cellularmonitor/blob/980ad15f70f4246bc2b749429d8b968c84aa7a74/src/components/SidePanel/__tests__/TraceFileInformation.test.tsx#L20-L26).

Also: Export a function `describeError` to convert any error object to a reasonable string representation. Because the previously [often used `error.message || error`](https://github.com/NordicSemiconductor/pc-nrfconnect-toolchain-manager/blob/df7bd77a53f1c1981b050a31b14e42fb64fcd3fe/src/Manager/Environment/effects/installToolchain.ts#L64-L65) is ugly to use with TypeScript and has some weaknesses (e.g. when `error` might be `null`, is an instance of `Error` or is an object which does not contain a `message` property).

Some additional, small cleanups: 5dce146, eaad2e4, c4b179b


src/logging/describeError.ts
```diff
@@ -35,9 +35,9 @@ const describeOrigin = (origin?: MaybeWithMessageAndOrigin) => {
     return ` (Origin: ${describe(origin)})`;
 };
 
-export default (message: string, error: unknown) => {
+export default (error: unknown) => {
     if (error == null) {
-        return message;
+        return String(error);
```
nicolai-nordic, line 40: Why are we casting `null` to a string here? Could we return something more explicit?

nicolai-nordic (approved):
Lots of really nice changes!

src/logging/describeError.ts
datenreisender: In this case the error can be either `null` or `undefined` (notice the double, not triple equals) which are converted to the strings `'null'` or `'undefined'`. 

Usually `null` or `undefined` should not be passed to this function but if it happens I thought it might be helpful to just show which one was passed.

PR #244, opened 2021-12-14
Fix/system report and types

PascalPickel:
- System report handles device object being null
- Website property formatted correctly in system report
- RangeProp type does not require decimals property


jonasem (approved):
Good work :)

PR #250, opened 2021-12-22
Feat/nrfdl log

PascalPickel:


src/Log/logSlice.ts
```diff
@@ -48,10 +52,16 @@ const slice = createSlice({
         toggleAutoScroll: state => {
             state.autoScroll = !state.autoScroll;
         },
+        toggleExtendedLogging: state => {
+            state.extendedLogging = !state.extendedLogging;
+            if (state.extendedLogging)
+                setLogLevel(getDeviceLibContext(), 'NRFDL_LOG_TRACE');
+            else setLogLevel(getDeviceLibContext(), 'NRFDL_LOG_CRITICAL');
```
datenreisender, line 59: It is strange that the first `setLogLevel` is on a line by itself, while the other is directly after the `else`. Please add a line break after `else`.

Please also extract this into a function and call that also at https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/b94bc284299c0c76dee885295d58b39f9011f52f/src/Log/logListener.ts#L74, so we only have the information, that `NRFDL_LOG_CRITICAL` is the default log level in one place.

src/Log/logListener.ts
```diff
@@ -63,13 +71,23 @@ const addLogEntriesToStore = (dispatch: TDispatch) => () => {
 function startListening(dispatch: TDispatch) {
     sendInitialMessage();
 
+    setLogLevel(getDeviceLibContext(), 'NRFDL_LOG_CRITICAL');
+    const taskID = startLogEvents(
+        getDeviceLibContext(),
+        err => {
+            if (err) logger.info(err.message);
```
datenreisender, line 78:
```suggestion
            if (err) logger.logError('Error while stopping log messages from nrf-device-lib', err);
```

src/Log/logListener.ts
```diff
@@ -52,6 +58,8 @@ const addLogEntriesToStore = (dispatch: TDispatch) => () => {
     }
 };
 
+const formatNrfdlLogs = (msg: string) => msg.split(' ').slice(1).join(' ');
```
datenreisender, line 61: If I understand this correctly, this essentially cuts of everything before the first space. Why is this done?

src/Log/logListener.ts
```diff
@@ -63,13 +71,23 @@ const addLogEntriesToStore = (dispatch: TDispatch) => () => {
 function startListening(dispatch: TDispatch) {
     sendInitialMessage();
 
+    setLogLevel(getDeviceLibContext(), 'NRFDL_LOG_CRITICAL');
+    const taskID = startLogEvents(
+        getDeviceLibContext(),
+        err => {
+            if (err) logger.info(err.message);
+        },
+        event => logger.debug(formatNrfdlLogs(event.message))
```
datenreisender, line 80: If I understand this correctly, here you always use log level `debug`. Would be better, if you use a log level fitting the log level of the log message. E.g. for `NRFDL_LOG_TRACE` and `NRFDL_LOG_DEBUG` use logger.debug, for `NRFDL_LOG_INFO` use `debug.info` etc.

datenreisender (changes requested):
Good work, some comments you should look at.

Also: Currently the default log level is `NRFDL_LOG_CRITICAL`. Sound pretty high and something like `NRFDL_LOG_WARNING` seems more appropriate. I think we usually want to see if device-lib logs a warning or at least if it logs an error. Or are there too many things that device-lib logs on warning or error level that we do not want to expose to our users?

src/Log/logSlice.ts
PascalPickel: Fixed

src/Log/logListener.ts
PascalPickel: Fixed

src/Log/logListener.ts
PascalPickel: Used that as a placeholder to remove the timestamp. Will keep the timestamp though and instead set a proper log pattern.

src/Log/logListener.ts
PascalPickel: Mapping the levels now.

Changelog.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 5.16.0 - 2022-01-07
+
+### Added
+
+-   Log messages from nRF-device-lib-js.
```
nicolai-nordic, line 13: Maybe a bit more verbose description? E.g. say that there is now an option to toggle nrfdl logging or something similar?

src/About/SupportCard.tsx
```diff
@@ -52,6 +70,39 @@ export default () => {
                     label="Create system report"
                 />
             </Section>
+            <Section title="Verbose Logging">
+                <p>
+                    Aid our support team with additional log information. Enable
+                    this only when necessary as the log will grow quickly.
```
nicolai-nordic, line 76: Small enhancement request, but maybe this text: `Enable this only when necessary as the log will grow quickly.` should be colored slightly different to make it stand out a bit as it is important that the users see this.

src/About/SupportCard.tsx
```diff
@@ -52,6 +70,39 @@ export default () => {
...
+                </p>
+                <Toggle
+                    id="enableVerboseLoggin"
+                    label="VERBOSE LOGGING"
+                    onToggle={() => {
+                        if (!verboseLogging)
```
nicolai-nordic, line 82: Hmm, maybe I'm just brainfarting here, but shouldn't this conditional be switched around? If `verboseLogging` is `true` we want to set TRACE-level logging right?

src/App/App.tsx
```diff
@@ -197,6 +205,25 @@ App.propTypes = {
...
+        const taskId = startLogEvents(
+            getDeviceLibContext(),
+            (err?: Error) => {
+                if (err)
+                    logger.logError(
+                        'Error while listening to log messages from nrf-device-lib',
+                        err
+                    );
+            },
+            event => logNrfdlLogs(event)
```
nicolai-nordic, line 219: [Very minor] This can be simplified to:

```suggestion
            logNrfdlLogs
```

src/About/SupportCard.tsx
PascalPickel: The condition gets toggled after after the if-else statement.

nicolai-nordic (approved):
Some minor comments, but all in all this looks really good and will be very useful!

src/About/SupportCard.tsx
nicolai-nordic: Can we swap the order of events then? Otherwise it's confusing, at least to me.

PR #306, opened 2022-01-25
Fix selected device in system report of `ErrorBoundary`

datenreisender:
When the app crashed and the error reporter was displayed, users can generate a system report. In that report, the current device was not displayed correctly. E.g. with a nRF52 Dongle with the RSSI app on this was shown:
```
- Current device:
  - name:          undefined
  - serialNumber:  [object Object]
  - cores:         undefined
  - website:       undefined
```

After this fix this is correctly shown like this:
```
- Current device:
  - name:          nRF52 RSSI Dongle
  - serialNumber:  F31282696FF5
  - cores:         Unknown
  - website:       Unknown
```


src/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -191,20 +187,9 @@ class ErrorBoundary extends React.Component<
     }
 }
 
-// @ts-ignore Don't know what to do with these yet
```
jonasem, line 194: Good riddance

src/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -80,14 +75,16 @@ class ErrorBoundary extends React.Component<
             ? sendUsageData(error.message)
             : sendGAEvent(error.message);
 
+        const selectedDevice =
```
jonasem, line 78: Maybe we should at some point consider changing devices into a `Map` to avoid syntax like this.

jonasem (approved):

src/ErrorBoundary/ErrorBoundary.tsx
datenreisender: After reading your comment I realised that this component directly reads the state instead of using selectors here: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/1f62d64ae751b9f6b78d4450144d44110aaa0ba0/src/ErrorBoundary/ErrorBoundary.tsx#L191-L192

If we switch to selectors, then this component does not really care whether the devices are internally a `Map` or something else. But doing that change proved to be more yak shaving when I realised that the `ErrorBoundary` is used in two different contexts: In apps, where we actually have devices, and in the launcher where there are no devices.

So I postponed doing that update until later and get the PR merged in it's current state.

PR #307, opened 2022-01-28
Update/nrfdl log dev

nicolai-nordic:
- Only propgate nrfdl log messages if in development (i.e. app is not packaged).
- Update nrfdl log message formatting slightly, i.e. add log-level.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 5.16.3 - 2022-01-28
+
+### Changed
+
+-   Only propagate device-lib logs for production builds.
+-   Update nrfdl log formatting.
+
 ## 5.16.2 - 2022-01-24
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "5.16.2",
+    "version": "5.16.3",
     "description": "Shared commodities for developing pc-nrfconnect-* packages",
     "repository": {
         "type": "git",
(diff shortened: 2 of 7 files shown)


Changelog.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 5.16.3 - 2022-01-28
+
+### Changed
+
+-   Only propagate device-lib logs for production builds.
```
datenreisender, line 13: The inverse is true, right? 
```suggestion
-   Only propagate device-lib logs while in development.
```

package.json
```diff
@@ -147,4 +147,4 @@
         "extends": "./config/eslintrc.json"
     },
     "prettier": "./config/prettier.config.js"
-}
+}
```
datenreisender, line 150: Please add the final newline back.

src/Device/deviceLister.ts
```diff
@@ -30,6 +31,7 @@ nrfDeviceLib.setTimeoutConfig(deviceLibContext, { enumerateMs: 3 * 60 * 1000 });
 export const getDeviceLibContext = () => deviceLibContext;
 
 export const logNrfdlLogs = (evt: LogEvent) => {
+    if (remote.app.isPackaged) return;
```
datenreisender, line 34: Nice, I did not know about `isPackaged`. Maybe we can use this also in other places instead of NODE_ENV.

datenreisender (commented):
With this changes in place, the logs from device-lib are _never_ propagated, and thus the “Verbose logging” does not work anymore, right? Is that intended?

nicolai-nordic:
> With this changes in place, the logs from device-lib are _never_ propagated, and thus the “Verbose logging” does not work anymore, right? Is that intended?

You are right, good catch, this is not intended behavior.

jonasem (approved):

PR #314, opened 2022-02-03
Show development as separate category in google analytics.

jonasem:
image


jonasem:
Also compact log file data of usage data
image

src/utils/usageData.ts
```diff
@@ -17,6 +17,9 @@ import {
 } from './persistentStore';
 
 const trackId = 'UA-22498474-5';
+const isDevelopment = process.env.NODE_ENV === 'development';
```
nicolai-nordic, line 20: This is the third way (at least) of checking if we are in development mode, I think we should agree on which we use :) I don't know which one is better, but we should definitely agree on what we want to use in the future.

Currently we are also using the `electron-is-dev` module in PPK and using `app.isPackaged` in `deviceLister.ts` also in shared.

nicolai-nordic (approved):
Other than landing on how we should check against dev environment it looks good to me!

src/utils/usageData.ts
jonasem: Sounds like a job for shared!

jonasem: Decided to go with app.isPackaged.

PR #322, opened 2022-02-14
Fix getting versions from `device-lib`

datenreisender:
Already before `device-lib` changed what it returns from `getModuleVersions`. This was adapted in `src/utils/logLibVersions.ts` but not in `src/utils/systemReport.ts` and `src/Log/logListener.ts`. This is fixed here.

Some additional changes:
- The whole code to retrieve a single module version is now encapsulated in the function `getModuleVersion` which also got unit tests.
- In the system report, the versions of `device-lib` and `device-lib-js` are also included.
- Remove an outdated hack in `src/utils/describeVersion.ts`
- Lower the log level of the `Application data folder` message.
- Bump version number

### DIFF
--- a/src/Device/deviceLibWrapper.ts
+++ b/src/Device/deviceLibWrapper.ts
@@ -5,12 +5,14 @@
  */
 
 import { app } from '@electron/remote';
-import nrfDeviceLib, {
+import {
+    createContext,
     Error,
     LogEvent,
     ModuleVersion,
     setLogLevel,
     setLogPattern,
+    setTimeoutConfig,
     startLogEvents,
     stopLogEvents,
 } from '@nordicsemiconductor/nrf-device-lib-js';
@@ -20,7 +22,7 @@ import logger from '../logging';
 import { RootState, TDispatch } from '../state';
 import { getVerboseLoggingEnabled } from '../utils/persistentStore';
 
-const deviceLibContext = nrfDeviceLib.createContext();
+const deviceLibContext = createContext();
 export const getDeviceLibContext = () => deviceLibContext;
 
 export const logNrfdlLogs =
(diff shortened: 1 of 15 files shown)


nicolai-nordic (approved):
Really nice, thumbs up for tests!

src/Device/deviceLibWrapper.ts
```diff
@@ -0,0 +1,97 @@
...
+                );
+        },
+        (evt: LogEvent) => dispatch(logNrfdlLogs(evt))
+    );
+    return () => {
+        stopLogEvents(taskId);
+    };
+};
+
+export const setDeviceLibLogLevel = (verboseLogging: boolean) =>
```
jonasem, line 70: I would expect, from the name, that the parameter would be the desired loglevel. If the name was something like setVerboseLogLevel or something, then the parameter makes more sense to me.

src/Device/deviceLister.ts
```diff
@@ -159,7 +116,7 @@ export const startWatchingDevices =
  * @returns {undefined}
  */
 export const stopWatchingDevices = () => {
-    // Start here
+    // Not sure, if this guard clause is really needed
```
jonasem, line 119: Seeing how the following command is wrapped in a try catch, it seems a bit unecceary? Also, I think checking if there is a hotplugTaskId seems like a better guard, and it would never exist without context.

jonasem (approved):

src/Device/deviceLibWrapper.ts
datenreisender: Hm, yes, maybe `setVerboseDeviceLibLogging` is slightly better. Will change this.

PR #325, opened 2022-02-23
Update list of available traits

nicolai-nordic:


jonasem:
It seems that the entire interface should be removed. The only place we actually use it, we cast it into a DeviceTraits from nrfdl-js. I suggest we use that instead.

nicolai-nordic:
> It seems that the entire interface should be removed. The only place we actually use it, we cast it into a DeviceTraits from nrfdl-js. I suggest we use that instead.

Good catch! Pushing update now.

jonasem (approved):

PR #330, opened 2022-03-03
Bundle shared with apps

jonasem:
When updating an app to this version, it will bundle shared. This requires the launcher used to also use this version of shared (or newer). This will let us update shared without updating the launcher in the future. 

This will increase the bundle size of the apps, as it bundles in shared.


datenreisender (approved):
🎉

PR #351, opened 2022-05-06
Reduce startup time of windows apps

jonasem:
Step one of reducing startup time. It's important to avoid eagerly importing 'systeminformation' on windows, as it will cause a exesync towards a "chcp" command which often takes several seconds. 

Next steps will be removing testing-library from the bundle and the pc-ble-driver, which in total will reduce loading about ~400ms


chunfantasy (approved):
LGTM

PR #362, opened 2022-05-16
Move testUtils out of main bundle

jonasem:
Importing testUtils from shared now has to use the path:
```
import { testUtils } from 'pc-nrfconnect-shared/test';
```
The problem with having it in the main index ts, was that it was loaded on starting apps at runtime.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -612,24 +612,6 @@ declare module 'pc-nrfconnect-shared' {
 
     // describeError.ts
     export const describeError: (error: unknown) => string;
-
-    export const testUtils: {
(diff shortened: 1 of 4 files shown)


PascalPickel (approved):

datenreisender:
This requires changes in all apps which currently use the `testUtils` when upgrading to the new version of `shared`. Will you do those changes? 

At least you should add `Changed` entry to the changelog that describes the change. Also consider a `Steps to upgrade when using this package`, especially if you do not plan to do all the needed changes yourself soon.

PR #372, opened 2022-06-27
Shortcuts integration

jesper2k:


src/Shortcuts/Shortcut-item.scss
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-4-Clause
+ */
+
+.shortcut-item {
+    margin-top: 10px;
+    .shortcut-span {
+        background-color: #cfd8dc;
```
jonasem, line 10: Use color defined in `styles.scss`

src/Shortcuts/ShortcutItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const shortcutKeys: string[] = hotKey.split('+');
+    return (
+        <Card className="shortcut-item">
+            <Card.Body>
+                <Card.Title>{title}</Card.Title>
+                <Card.Text>
+                    {shortcutKeys.map(shortcutKey => (
+                        <span key={shortcutKey}>
+                            <span className="shortcut-span">{shortcutKey}</span>
+                            {shortcutKey !==
```
jonasem, line 29: Alternatively use something like the index to check for last item, and use a short-circuiting && instead of a ternary.

jonasem: E.g. { a === b ? c : null } => { a === b && c }

src/Shortcuts/useHotkey.tsx
```diff
@@ -0,0 +1,37 @@
...
+            shortcuts.splice(
+                shortcuts.findIndex(
+                    shortcutItem => shortcutItem.hotKey === shortcut.hotKey
+                ),
+                1
+            );
+
+            Mousetrap.unbind(shortcut.hotKey);
+        };
+    }, [deps, shortcut]); // Sould this not just be "}, deps);"?
```
jonasem, line 36: I would think so, if the caller wanted shortcut to be in the deps, it could just put it there. Also, the deps-list is not spread, so this wouldn't work as the caller intended to either way.

src/Shortcuts/Shortcut-item.scss
jesper2k: Will go through design with Ketil before final PR

src/Slider/Handle.tsx
```diff
@@ -96,7 +96,9 @@ const Handle: FC<Props> = ({
             style={{ left: `${percentage}%` }}
             onMouseDown={disabled ? noop : grabHandle}
             role="slider"
-            tabIndex={disabled ? -1 : 0}
+            tabIndex={-1}
+            // tabIndex={disabled ? -1 : 0}
+            // Always -1 as handles shouldn't be tab-able
```
jesper2k, line 101: Will revert, as it's not relevant to this PR

jesper2k (commented):
Jesper and Line reviewed files, will make new commit with improvements.

Otherwise this is fantastic. 🎉

src/App/VisibilityBar.tsx
```diff
@@ -32,13 +33,49 @@ const VisibilityBar: FC<{ isSidePanelEnabled: boolean }> = ({
...
+        isGlobal: true,
+        action: () => dispatch(clear()),
+    });
+
+    useHotKey({
+        hotKey: 'shift+o',
+        title: 'Open log file',
+        isGlobal: true,
+        action: () => logger.openLogFile(),
+    });
```
aadnekar, line 55: I'm guessing you may have discussed with Ketil about what the shortcuts should be?
For me `shift + o` seems like a good idea for opening something, and it allows us to keep `ctrl + o` for opening external files in e.g. PPK.
Is the shift key reserved for global shortcuts?

aadnekar (approved):
Think it looks really good, really well done :+1:

src/App/VisibilityBar.tsx
jesper2k: Did some smalltalk, but will get a review of the current hotkeys and change if needed. That would be the last commit before merge (hopefully...)

jonasem:
Add a changelog entry and bump the version in package.json

src/Shortcuts/ShortcutModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+            <Modal.Title>
+                <h3>Shortcuts</h3>
+            </Modal.Title>
+        </Modal.Header>
+        <Modal.Body className="shortcut-lists">
+            <div>
+                <h4 className="list-header">In this app</h4>
+                {shortcuts
+                    .filter(shortcut => shortcut.isGlobal === false)
+                    .sort((s1, s2) => s1.title.localeCompare(s2.title))
```
jonasem, line 38: Rememeber, sort mutates, so it can even be done upon pushing it into the array.

src/utils/useHotKey.tsx
```diff
@@ -4,16 +4,39 @@
...
+import { sendUsageData } from './usageData';
+
+export interface Shortcut {
+    hotKey: string;
+    title: string;
+    isGlobal: boolean;
+    action: () => void;
+}
+
+export const shortcuts: Shortcut[] = [];
```
datenreisender, line 19: Instead of using an array for `shortcuts` it is probably easier to use a `Map`, indexed by the hotkey. With this, removing an entry is just a simple function call instead of having to use `splice`

src/utils/useHotKey.tsx
```diff
@@ -4,16 +4,39 @@
...
+
+export default (shortcut: Shortcut, deps?: DependencyList) => {
```
datenreisender, line 21: Exposing the `DependencyList` from `useEffect` is wrong. Just remove the parameter. The way it is used in the RSSI app when calling `useHotKey` is AFAICS not needed at all.

src/App/VisibilityBar.tsx
```diff
@@ -32,13 +33,49 @@ const VisibilityBar: FC<{ isSidePanelEnabled: boolean }> = ({
     const isLogVisible = useSelector(isLogVisibleSelector);
     const autoScroll = useSelector(autoScrollSelector);
 
+    useHotKey({
+        hotKey: 'shift+m',
+        title: 'Show/Hide menu',
```
datenreisender, line 38: I think calling this “menu” is confusing because it isn't a side panel, not really a menu. I know that we also call it “menu” in the bar at the bottom and we should probably talk to Ketil whether we also rename it there. At least I expect something very different when it comes to a “menu” (e.g. a menu bar).

src/App/VisibilityBar.tsx
```diff
@@ -32,13 +33,49 @@ const VisibilityBar: FC<{ isSidePanelEnabled: boolean }> = ({
     const isLogVisible = useSelector(isLogVisibleSelector);
     const autoScroll = useSelector(autoScrollSelector);
 
+    useHotKey({
+        hotKey: 'shift+m',
```
datenreisender, line 37: You are using a wild mix of modifiers (sometimes shift, sometimes alt, sometimes ctrl). Especially using shift alone as a modifier for a shortcut is rather uncommon (shift+m alone usually just means “Capital M”). Could you maybe limit this to using only ctrl and alt?

datenreisender (commented):
Nice work, that contribution is helpful, thanks for that!

I think there is one urgent thing to address: You are currently breaking an existing API, as I comment in `index.d.ts`. 

The other comments are only “nice to haves”. They would make the shortcuts even better but it is up to you whether you want to address them.

Some of these are UX suggestions, so it is wise to discuss these also with Ketil. He is more knowledgable about UX than I am.

I like the visual representation of the keys.

Please excuse that I was not fast enough to take a look at this before you merged it (not your fault, you had it already open for quite some time).

Again: Well done!

src/NavBar/NavMenu.tsx
```diff
@@ -5,17 +5,31 @@
...
+    const dispatch = useDispatch();
+
+    useHotKey({
+        hotKey: 'ctrl+tab',
+        title: 'Switch pane',
+        isGlobal: true,
+        action: () => {
+            currentPane === paneNames.length - 1
+                ? dispatch(setCurrentPane(0))
+                : dispatch(setCurrentPane(currentPane + 1));
```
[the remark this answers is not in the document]
jesper2k: Me and Line discussed this and preferred the ternary. We can change to remainder function.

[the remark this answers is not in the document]
datenreisender: No, leaving it is fine.

src/Device/DeviceSelector/DeviceSelector.tsx
```diff
@@ -93,6 +94,13 @@ const DeviceSelector: FC<Props> = ({
 
     useAutoselectDevice(doSelectDevice);
 
+    useHotKey({
+        hotKey: 'alt+s',
+        title: 'Select device',
+        isGlobal: true,
+        action: () => toggleDeviceListVisible(),
+    });
```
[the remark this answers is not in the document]
jesper2k: We were actually working on this before we decided to finish the PR, but decided together with @jonasem that it could be a later feature, as we have already spent a lot of time on this project and have another (large one) to finish before summer ends.

But we agree, this will be the next feature if we finish the next project early.

src/App/VisibilityBar.tsx
```diff
@@ -32,13 +33,49 @@ const VisibilityBar: FC<{ isSidePanelEnabled: boolean }> = ({
...
+        isGlobal: true,
+        action: () => logger.openLogFile(),
+    });
+
+    useHotKey({
+        hotKey: 'shift+a',
+        title: 'Autoscroll log',
+        isGlobal: true,
+        action: () => dispatch(toggleAutoScroll()),
+    });
```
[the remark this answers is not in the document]
jesper2k: I agree on "autoscroll" and "open log file", but "Clear log" seems to be nice to have in my opinion 😸

src/utils/useHotKey.tsx
jesper2k: This was one of the first things we discussed when we started implementing shortcuts, but I can't for the life of me remember why we ended up using an array. It may have had something to do with being able to add more shortcuts on the same key, but I agree, this seems like the preferred way to do it.

src/utils/useHotKey.tsx
jesper2k: But we are re-rendering the same button depending on the `isPaused` bool which is being paused here. Removing this dependencies-paremeter wouldn't give us a way to update the shortcuts-array with a toggle-button since it would need to be two components instead, no?

We added the dependency-array to fix this issue.

src/App/VisibilityBar.tsx
jesper2k: Discussed this with @ketile, who said that it's not really a problem. We will add shortcuts in the apps tomorrow, so any suggestions on which shortcuts bind to which actions is appreciated 👍 

It's difficult to make shortcuts for "Save", "Select" and "Start/Stop" without one of them needing to use uncommon keys, though "Start/Stop" may be better to bind to spacebar.

src/App/VisibilityBar.tsx
jesper2k: We just saw "menu" on the button and used that name 🤷

src/utils/useHotKey.tsx
datenreisender: You actually have code, that depends on the assumption, that there are no two shortcuts with the same key:

* This might remove the wrong shortcut if there are two with the same shortcut: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/6c174f450f993f6d984e574f2ae9870fd22cad2e/src/utils/useHotKey.tsx#L31-L36
* Here the same key would be used for React if there are two with the same shortcut:  https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/6c174f450f993f6d984e574f2ae9870fd22cad2e/src/Shortcuts/ShortcutModal.tsx#L53-L57

src/utils/useHotKey.tsx
datenreisender: AFAICS the `isPaused` there does not have any effect. Which effect do you want to achieve? 

If we want to enable users of the hook to update the shortcut then we probably should put `shortcut` in the dependencies of `useEffect`. But that would have some consequences so I first want to understand what you need this for.

datenreisender:
One more thing: The previous version of `useHotKey` made it possible to pass an array of shortcuts, so that they all cause the same effect. You removed that. 

It is currently only used in the launcher, so it is not critical. Because there we do not have the shortcut modal yet and we can instead call `useHotKey` multiple times. But still it would be nice if that was also possible with the new `useHotKey`.

PR #402, opened 2022-09-14
feat: add @babel/plugin-transform-classes

PascalPickel:


datenreisender:
Why is the babel plugin `@babel/plugin-transform-classes` needed?

PascalPickel:
Was necessary for pc-xterm-lib

PR #404, opened 2022-09-23
Bundle publish script

jonasem:
In order to use this, when upgrading shared in an app, replace the publish script like this:

`"nordic-publish": "./dist/nordic-publish.js",`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "6.6.2",
+    "version": "6.6.3",
     "description": "Shared commodities for developing pc-nrfconnect-* packages",
     "repository": {
         "type": "git",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "bin": {
         "nrfconnect-license": "./scripts/nrfconnect-license.ts",
         "check-for-typescript": "./scripts/check-for-typescript.ts",
-        "nordic-publish": "./scripts/nordic-publish.ts"
+        "nordic-publish": "./scripts/nordic-publish.js"
     },
     "main": "src",
     "scripts": {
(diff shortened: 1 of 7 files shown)


scripts/nrfconnect-license.ts
```diff
@@ -142,7 +142,11 @@ const allSourceFiles = () =>
             '*.sass',
         ],
         {
-            ignore: [...entriesInGitignore, '.git'],
+            ignore: [
+                ...entriesInGitignore,
+                'scripts/nordic-publish.js',
```
aadnekar, line 147: Is this added because the license is not put in the very top of the file?

aadnekar (approved):
Looks good to me :+1:

scripts/nrfconnect-license.ts
jonasem: There is code `"use strict"` before the block with the license. So yes. Might alternatively change the script to look through the entire file for the license.

package.json
```diff
@@ -20,7 +20,8 @@
         "check:lint": "eslint --color .",
         "check:types": "ts-node scripts/check-for-typescript.ts tsc --noEmit --pretty",
         "check:license": "ts-node scripts/nrfconnect-license.ts check",
-        "prepare": "husky install"
+        "prepare": "husky install",
+        "update-publish-script": "tsc scripts/nordic-publish.ts -t es2015 -m node16"
```
datenreisender, line 24: We could probably also use `swc` or `esbuild` instead of `tsc`. But this compilation is so fast and done so rarely that it is fine to just use `tsc`.

scripts/nrfconnect-license.ts
datenreisender: We could also change the script to also skip the `"use strict"`. But so far, this is the only exception and it is fine to just exclude it this way here.

datenreisender (approved):

PR #409, opened 2022-10-24
Feature/add new functionality to dropdown

kylebonnici:
Add:

- Opt-in scroll in case the number of dropdown items is too much
- Add External Updating

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -38,7 +38,7 @@ export interface DropdownProps {
     disabled?: boolean;
     defaultIndex?: number;
     selectedItem?: DropdownItem;
-    maxHeight?: string;
+    noItemsBeforeScroll?: number;
 }
 
 const Dropdown: React.FC<DropdownProps> = ({
@@ -48,7 +48,7 @@ const Dropdown: React.FC<DropdownProps> = ({
     disabled = false,
     defaultIndex = 0,
     selectedItem = items[defaultIndex],
-    maxHeight = '',
+    noItemsBeforeScroll = 0,
 }) => {
     const [selected, setSelected] = useState(items[defaultIndex]);
     const [isActive, setIsActive] = useState(false);
@@ -94,8 +94,15 @@ const Dropdown: React.FC<DropdownProps> = ({
                 <img className={styles.image} src={chevron} alt="" />
             </button>
             <div
-                style={{ maxHeight }}
-                data-height={maxHeight.length > 0}
+                style={
+                    noItemsBeforeScroll > 0
+                        ? { maxHeight: `${noItemsBeforeScroll * 24}px` }
+                        : {}
+                } // 24px from Dropdown.module.scss .item height
+                data-height={
+                    noItemsBeforeScroll > 0 &&
+                    items.length > noItemsBeforeScroll
+                }
                 className={`${styles.content} ${
                     isActive ? styles.itemsActive : styles.itemsInactive
                 }`}
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -50,7 +50,9 @@ const Dropdown: React.FC<DropdownProps> = ({
     selectedItem = items[defaultIndex],
     noItemsBeforeScroll = 0,
 }) => {
-    const [selected, setSelected] = useState(items[defaultIndex]);
+    const [selected, setSelected] = useState(
+        items[items.findIndex(e => e.value === selectedItem.value)]
+    );
     const [isActive, setIsActive] = useState(false);
 
     useSynchronisationIfChangedFromOutside<DropdownItem>(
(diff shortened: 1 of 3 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -4,25 +4,41 @@
...
 
-interface DropdownProps {
+export interface DropdownProps {
     label?: string;
     items: DropdownItem[];
     onSelect: (item: DropdownItem) => void;
     disabled?: boolean;
     defaultIndex?: number;
+    selectedItem?: DropdownItem;
+    noItemsBeforeScroll?: number;
```
PascalPickel, line 41:
```suggestion
    ItemCountBeforeScroll?: number;
```
I kept reading it as "no items before scroll" even after realising it was "number ..."
Not sure if my naming suggestion is anymore intuitive though...

src/Dropdown/Dropdown.tsx
```diff
@@ -65,10 +90,23 @@ const Dropdown: React.FC<DropdownProps> = ({
                 onClick={onClick}
                 disabled={disabled}
             >
-                <span>{selected?.label}</span>
+                <span>
+                    {items.findIndex(e => e.value === selected.value) === -1
+                        ? ''
+                        : selected.label}
```
PascalPickel, line 96: Unless I misunderstand, this is simply to ensure the value is a valid one?
I don't see a scenario where that could be the case though.
Could you explain?

This can be shortened a bit either way (just personal preference)
`items.find(e => e.value === selected.value) ? '' : selected.label`

PascalPickel (commented):
Great stuff removing triggers on selecting the currently selected value

src/Dropdown/Dropdown.tsx
```diff
@@ -31,10 +47,19 @@ const Dropdown: React.FC<DropdownProps> = ({
     onSelect,
     disabled = false,
     defaultIndex = 0,
+    selectedItem = items[defaultIndex],
```
PascalPickel, line 50: It isn't very clear that this is simply just the first selected item. I.e. "defaultItem"

You could consider removing the defaultIndex and only allow a defaultItem even though it breaks API.

src/Dropdown/Dropdown.tsx
kylebonnici: how about num to keep it short?

src/Dropdown/Dropdown.tsx
kylebonnici: Given that the data is set by the parent component, there might be the case that the user enter a new DropownItem object such as {value, key} and:

- Since it is a different obj from the on in the items we can only match the value to ensure that it is the same one
- Since it is a different obj it can not even be in the list of items

Hence I felt the need to do data validation.

src/Dropdown/Dropdown.tsx
kylebonnici: Yes I keep it there not to break other apps + I think if one it not interested in external updates then they can just use that. but I know this creates a paradox if both a set.... so not sure how to proceed.

src/Dropdown/Dropdown.tsx
PascalPickel: Not sure if you are genius or if I'm blind. I think that's perfect

src/Dropdown/Dropdown.tsx
PascalPickel: Yeah, I think it's fair to force the usage of one. I remember wanting to create a default based on the item but the index was more convenient at the time.
Item is more explicit and verbose though so I'd prefer that.

src/Dropdown/Dropdown.tsx
```diff
@@ -31,10 +47,19 @@ const Dropdown: React.FC<DropdownProps> = ({
...
+    noItemsBeforeScroll = 0,
 }) => {
-    const [selected, setSelected] = useState(items[defaultIndex]);
+    const [selected, setSelected] = useState(
+        items[items.findIndex(e => e.value === selectedItem.value)]
```
PascalPickel, line 54:
```suggestion
        items.find(e => e.value === selectedItem.value)
```

PascalPickel (approved):

datenreisender:
Please describe the change in `Changelog.md`.

src/Dropdown/Dropdown.tsx
```diff
@@ -65,10 +90,23 @@ const Dropdown: React.FC<DropdownProps> = ({
...
+                        : selected.label}
+                </span>
                 <img className={styles.image} src={chevron} alt="" />
             </button>
             <div
+                style={
+                    noItemsBeforeScroll > 0
+                        ? { maxHeight: `${noItemsBeforeScroll * 24}px` }
+                        : {}
+                } // 24px from Dropdown.module.scss .item height
```
datenreisender, line 105: Instead of hard coding and explaining in a comment the `24`, can you instead export the value from the SCSS (probably using an [`:export` pseudo-selector](https://github.com/css-modules/icss#export)) and use that in here?

PR #417, opened 2022-11-04
Add external updating for state selector

kylebonnici:


PascalPickel (approved):

datenreisender:
This makes the StateSelector act as either a controlled or an uncontrolled component. Supporting both often makes the code more complicated and quickly introduces subtle bugs, at least for corner cases.

If somehow possible I usually suggest sticking to one of the two options, most of the times a controlled component is the better way, as is also mentioned in the React docs. https://reactjs.org/docs/glossary.html#controlled-vs-uncontrolled-components

So, especially because this is a new component that is not widely used yet, can you please reconsider this design choice?

PR #424, opened 2022-11-14
Extend slider and inlineinput

kylebonnici:


src/Slider/slider.scss
```diff
@@ -9,6 +9,6 @@
 .slider {
     height: 2*$handle-radius-selected;
 
-    position: relative;
+    position: relative;
```
jonasem, line 12: Guessing this wasn't entirely intentional ;)

src/Slider/percentage.ts
```diff
@@ -12,12 +12,51 @@ export const constrainedToPercentage = (percentage: number) => {
...
     value: number,
-    { min, max, decimals = 0 }: RangeProp
-) => Number(((value * (max - min)) / 100 + min).toFixed(decimals as number));
+    { min, max, decimals = 0, step = null, explicitRange = [] }: RangeProp,
+    directionForward: boolean
+) => {
+    if (explicitRange.length > 0) {
+        const noOfIndexes = explicitRange.length - 1;
+        const computedValue = Number(
+            ((value * (max - min)) / 100 + min).toFixed(decimals as number)
```
jonasem, line 27: A lot of `decimals as number` in this file, stemming from the proptypes object I guess. We've previously discussed just getting rid of proptypes all together now that everything is rewritten i typescript. Maybe we should touch in on that discussion again soon.

src/Slider/Ticks.tsx
```diff
@@ -20,16 +20,22 @@ interface Props {
 
 const Ticks: FC<Props> = ({
     valueRange,
-    range: { min, max, decimals = 0 },
+    range: { min, max, decimals = 0, step = null, explicitRange = [] },
```
aadnekar, line 23: Do we need to have a default value for `step` and `explicitRange`, can't we treat `undefined` and `null` as equals, in the case for `step` at least.

src/Slider/percentage.ts
datenreisender: There are definitely too many type casts in here. 

I do not think, that it is because of the PropTypes. E.g. in this case here the TypeScript type for `RangeProp` says, that decimals can be `undefined` and `null`: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/511c3a779575d29aec152f9df331d70a0a70a7ef/src/Slider/rangeShape.ts#LL12 

So, effectively, the `decimals as number` here is just a non-null assertion. That would have been better written by using the `!` operator for non-null-assertions as `decimals!`. But even that seems bad here, because readers have to investigate the types really well (and be knowledgable about the coercion rules good to enough, to know that a `null` will be turned into a `0` in this place).

I am not familiar enough with this component anymore. First: Is it really intended, that `decimals` can be, besides ` number`, be either `null` or `undefined`? Why are both needed? I suggest to constrain this to only one of the two.

Looking at the code here I assume, that if `decimals` is either `null` or `undefined`, that it should be interpreted as `0`, right? Then I suggest, that this is made explicit just once, e.g. in the `Slider` component. And that all other functions and places work with a `RangeProp` type where `decimals` cannot be `null` anymore but will always contain a number. E.g. at the top of `Slider` we could do something like this:

```ts
const Slider: FC<Props> = ({
    // ...
    range: rangeWithOptionalValues,
    // ...
}) => {
    const range = {
        ...rangeWithOptionalValues,
        decimals: rangeWithOptionalValues.decimals ?? 0
    };
    // ...
```

I did not look close enough at the code to get an idea whether something similar also makes sense for `steps`. But I strongly suggest that we also do not allow both `undefined` and `null` for `steps`, unless there is a clear need for it.

@Kyle: I am aware, that you did not start with those `x as number` here, they were already in this code before, you just copied that bad code. Sorry, but we should not let that grow. The same is true for defining the property to be either `null` or `undefined`. That was already there for `decimals` and you did the same for `steps`.

datenreisender:
What is the intention of the PR? The title says “Extend slider and inlineinput” but in what way should it extend it the two? Please mention that in the description of the PR. 

If you think the commit messages explain that intention well enough, then mention this fact in the description of the PR and I will read the commit messages instead. But a little summary beyond the title of the PR would be nice.

PR #425, opened 2022-11-16
Transpile publish script on install

datenreisender:
When this package is installed, `scripts/nordic-publish.ts` is transpiled and bundled. This has several advantages over the previous, pre-transpiled variant:

- We do not need to remember ot update scripts/nordic-publish.js.
- The previous version was only transpiled, not bundled. Which had the problem, that e.g. wrong versions of used packages, like `commander`, could be used.

Even better would be to transpile the script automatically, when publishing a new version of shared, but we do not publish real packages of shared yet.


package.json
```diff
@@ -21,7 +21,7 @@
         "check:types": "ts-node scripts/check-for-typescript.ts tsc --noEmit --pretty",
         "check:license": "ts-node scripts/nrfconnect-license.ts check",
         "prepare": "husky install",
-        "update-publish-script": "tsc scripts/nordic-publish.ts -t es2015 -m node16"
+        "postinstall": "esbuild scripts/nordic-publish.ts --bundle --outfile=scripts/nordic-publish.js --platform=node --log-level=warning"
```
jonasem, line 24: Changing to esbuild for performance reasons?

Changelog.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+-   `scripts/nordic-publish.js` is now transpiled and bundled on install, so we
```
jonasem, line 13: Did you intend to delete the file from the git repo?

scripts/nordic-publish.js
```diff
@@ -1,210 +1,8268 @@
-#!/usr/bin/env node
+#!/usr/bin/env ts-node
```
jonasem, line 1: I guess we still have to change this to node?

package.json
datenreisender: Changing because I want to bundle all dependencies (`commander`, `semver`, `shasum`) instead relying that they are present on the machine doing the install in compatible versions.

And `tsc` can only transpile but cannot bundle.

Changelog.md
datenreisender: Yes, that was my intention but I failed to do so. :-) Seems I messed it up because I also added it to `.gitignore`. Thanks for noticing, fixed it in 39f87c8179.

scripts/nordic-publish.js
datenreisender: You are right. But we cannot change this manually any longer when we generate it during installing the package. :-/ 

I could add a `transform` to the webpack configuration that copies the script, but I think an even easier solution is, if we call the script explicitly with node. This would also mean, that we do not need to set the executable bit on the file any longer. So in `package.json` we would have `"nordic-publish": "node ./dist/nordic-publish.js"`. 

Sounds good? I updated the instructions in `Changelog.md` in 64b98ad.

jonasem: Yeah, my other thought would be to rewrite `ts-node` to `node` in the ts-file, since it probably won't be run directly anymore.

PR #431, opened 2022-11-20
Feature/add esbuild bundler

jonasem:


package.json
```diff
@@ -41,6 +42,7 @@
         "@mdi/font": "7.0.96",
         "@nordicsemiconductor/nrf-device-lib-js": "^0.4.13",
         "@reduxjs/toolkit": "1.8.3",
+        "@svgr/core": "^6.5.1",
```
chunfantasy, line 45: Same as in the launcher, shall we pin them?

chunfantasy (approved):
LGTM! Very great work Jonas!

package.json
jonasem: Good point, will do it when I remove webpack.

PR #446, opened 2022-12-01
Enhance specifying ranges

datenreisender:
I was unhappy about the introduction of so called `explicitRange`s. 

A range like this is easily possible but obviously an error:
```ts
{
  min: 3,
  max: 6,
  explicitRange: [4, 6, 7, 8]
}
```

Initially not clear to me was that also a range like this leads to buggy behaviour:
```ts
{
  min: 3,
  max: 6,
  explicitRange: [3, 3.1, 6]
}
```

That second example is missing appropriate values for the optional properties `step` and `decimals`. 

But, even more important, when specifying explicit values, the other range properties should not really be necessary, because they can be deduced. 

So this PR changes that a range can be _either_ specified by enumerating the values _or_ by specifying the properties `min`, `max`, and optionally `step` and `decimals`. 

This is valid for both components where this applies: `Slider` and `NumberInlineInput`. So a `NumberInlineInput` element can now be created in one of these two ways:

```jsx
<>
  <NumberInlineInput
    values={[4]}
    range={[3, 4, 6, 7]}
    onChange={[() => {}]}
  />

  <NumberInlineInput
    values={[4]}
    range={{ min: 3, max: 7 }}
    onChange={[() => {}]}
  />
</>
```

Because TypeScript in many cases ignores excessive properties, it sometimes does not raise a type error when still using the `explicitRange` property. But that property is not supported anymore, so there is also a runtime check for it and an error is printed if it is still used.

### Other changes

- Like before it is not implemented to show ticks in a `Slider` when using explicit values. But before we showed the ticks in wrong locations and a warning. Now the ticks are not shown at all in this case and a slightly different warning is shown. (implemented as part of e78309529)
- 8bc8f35: Fixed: Sliders with ticks without a `step` were rendered incorrectly, e.g. 
    ```jsx
  <Slider
       ticks
       values={[2]}
       range={{ min: 0, max: 5 }}
       onChange={[() => {}]}
   />
    ```
- 4bf1185: The properties of a range were previously checked for consistency on every render. Now they are only checked in the beginning and when they are changed.
- b78b336: The properties of a range were previously checked for consistency for a `Slider`, now they are also checked for a `NumberInlineInput`. 
- 547cf34: Check some more properties for consistency: That `decimals` (when specified) must be a non-negative integer and that `step` (when specified) must be larger than zero.
- e4dd35f: Before `min` and `max` needed to be multitudes of `step`. But it is actually sufficient if only the difference between `max` and `min` is a multitude of `step`.

### Remaining aspects

- With a range like `{min: 1, max: 7, step: 0.5}` no warning is printed, but `Slider` (or `NumberInlineInput`) act buggy, unless also `decimals: 1` is specified. But maybe we should not try to catch and warn about all possible errors with properties.

### DIFF
--- a/src/InlineInput/NumberInlineInput.tsx
+++ b/src/InlineInput/NumberInlineInput.tsx
@@ -7,7 +7,13 @@
 import React, { FC } from 'react';
 
 import { isFactor } from '../Slider/factor';
-import { isValues, Range, RangeOrValues, Values } from '../Slider/range';
+import {
+    isValues,
+    Range,
+    RangeOrValues,
+    useValidatedRangeOrValues,
+    Values,
+} from '../Slider/range';
 import InlineInput from './InlineInput';
 
 import './number-inline-input.scss';
@@ -79,21 +85,25 @@ const NumberInlineInput: FC<Props> = ({
     range,
     onChange,
     onChangeComplete = () => {},
-}) => (
-    <InlineInput
-        className="number-inline-input"
-        disabled={disabled}
-        value={String(value)}
-        isValid={newValue => isValid(Number(newValue), range)}
-        onChange={newValue => onChange(Number(newValue))}
-        onChangeComplete={newValue => onChangeComplete(Number(newValue))}
-        onKeyboardIncrementAction={() =>
-            changeValueStepwise(value, range, 1, onChange)
-        }
-        onKeyboardDecrementAction={() =>
-            changeValueStepwise(value, range, -1, onChange)
-        }
-    />
-);
+}) => {
+    useValidatedRangeOrValues(range);
+
+    return (
+        <InlineInput
+            className="number-inline-input"
+            disabled={disabled}
+            value={String(value)}
+            isValid={newValue => isValid(Number(newValue), range)}
+            onChange={newValue => onChange(Number(newValue))}
+            onChangeComplete={newValue => onChangeComplete(Number(newValue))}
+            onKeyboardIncrementAction={() =>
+                changeValueStepwise(value, range, 1, onChange)
+            }
+            onKeyboardDecrementAction={() =>
+                changeValueStepwise(value, range, -1, onChange)
+            }
+        />
+    );
+};
 
 export default NumberInlineInput;
--- a/src/InlineInput/NumberInlineInput.tsx
+++ b/src/InlineInput/NumberInlineInput.tsx
@@ -8,6 +8,7 @@ import React, { FC } from 'react';
 
 import { isFactor } from '../Slider/factor';
 import {
+    getStep,
     isValues,
     Range,
     RangeOrValues,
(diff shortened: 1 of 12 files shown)


src/Slider/range.ts
```diff
@@ -0,0 +1,89 @@
...
+export const getMin = (rangeOrValues: RangeOrValues) =>
+    isValues(rangeOrValues) ? rangeOrValues[0] : rangeOrValues.min;
+
+export const getMax = (rangeOrValues: RangeOrValues) =>
+    isValues(rangeOrValues)
+        ? rangeOrValues[rangeOrValues.length - 1]
+        : rangeOrValues.max;
+
+export const getStep = (range: Range) =>
+    range.step != null ? range.step : 0.1 ** (range.decimals ?? 0);
```
kylebonnici, line 34: Consider to avoid floating point number issues: 
```suggestion
    range.step != null ? range.step : Number(0.1 ** range.decimals ?? 0).toFixed(range.decimals ?? 0);
```

As 

`Number(0.1 ** 3).toFixed(3) = 0.001`

while

`0.1 ** 3 = 0.0010000000000000002`

src/InlineInput/NumberInlineInput.tsx
```diff
@@ -7,63 +7,72 @@
...
+
+const isInValues = (value: number, values: Values) => values.includes(value);
+
+const isInRange = (value: number, { min, max, decimals, step }: Range) =>
     value >= min &&
     value <= max &&
     value === Number(value.toFixed(decimals)) &&
-    (step > 0 ? isFactor(value, step) : true) &&
-    (explicitRange.length > 0 ? explicitRange.indexOf(value) !== -1 : true);
+    (step == null ? true : isFactor(value, step));
```
kylebonnici, line 36: to fix the issue for `{min: 1, max: 7, step: 2}` in 'NumberInlineInput' we need to do the below here 

```suggestion
    (step == null ? true : isFactor(value - min, step));
```

This shift the range to make use of isFactor

src/Slider/range.ts
```diff
@@ -0,0 +1,89 @@
...
+            `range.step must be larger than zero but is ${range.step}`
+        );
+        assert(
+            isFactor(range.min, range.step),
+            `range.step must be a factor of range.min: ${JSON.stringify(range)}`
+        );
+        assert(
+            isFactor(range.max, range.step),
+            `range.step must be a factor of range.max: ${JSON.stringify(range)}`
+        );
```
kylebonnici, line 77:
```suggestion
        assert(
            isFactor(range.max - range.min, range.step),
            `range.step must be a factor of range.max: ${JSON.stringify(range)}`
        );
```

kylebonnici:
Nice work !! I like it a lot more and it is more programmer friendly like this

kylebonnici (approved):

src/InlineInput/NumberInlineInput.tsx
```diff
@@ -7,63 +7,72 @@
...
-        const newIndex = currentIndex + steps;
+    return values[newIndex];
+};
 
-        if (newIndex >= 0 && newIndex < range.explicitRange.length) {
-            action(range.explicitRange[newIndex]);
-        }
+const nextInRange = (current: number, range: Range, steps: number) => {
+    const newValue = Number(
+        (current + steps * getStep(range)).toFixed(range.decimals)
```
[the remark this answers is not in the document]
kylebonnici: We discussed this the developer needs to make sure that  `step` and `decimals` are set correctly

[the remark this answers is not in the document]
datenreisender: If decimals is not set, `toFixed()` treats it as `0`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#syntax

PR #447, opened 2022-12-01
Adapt SerialPort wrapper to changes in launcher

aadnekar:
When a renderer writes to the port, the renderer should wait for the onDataWritten callback before writing the data to the terminal. This is in order to synchronize the data with other renderers.

A renderer can also provide a `settingsLocked` option in order to avoid others to overwrite the settings. The overwrite flag is no longer a single argument but is combined with settingsLocked into an object.


kylebonnici (approved):
Looks ok to me :)

PR #448, opened 2022-12-01
Enforce use of unknown in catch blocks

aadnekar:
The setting will automatically give the variable type unknown, similar to `catch (error: unknown) {}`.

The main reason to do this is in order to not accidentally throw something that is not an Error, because that would not provide a stack trace, and will therefore make it more difficult to debug.

In order to check if the variable is an error:
    if (error instanceof Error) throw err
    else throw new Error(`${error}`)

### DIFF
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -10,6 +10,8 @@
         "strict": true,
         "isolatedModules": true,
         "forceConsistentCasingInFileNames": true,
-        "resolveJsonModule": true
+        "resolveJsonModule": true,
+
+        "useUnknownInCatchVariables": true
     }
 }


datenreisender:
We already have `strict` turned on: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/5f9c2cda1f18e99cc53a9b408730e51b32fd3ada/config/tsconfig.json#LL10

With `strict` turned on, this is enabled by default: https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

So I think there is no need to turn it on explicitly again. Or did you encounter a case, where turning it on explicitly does change something?

aadnekar:
> We already have `strict` turned on: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/5f9c2cda1f18e99cc53a9b408730e51b32fd3ada/config/tsconfig.json#LL10
> 
> With `strict` turned on, this is enabled by default: https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables
> 
> So I think there is no need to turn it on explicitly again. Or did you encounter a case, where turning it on explicitly does change something?

Did not know it was on if `strict` was enabled. I'll close this then ☺ Thanks @datenreisender

PR #453, opened 2022-12-12
New Dialog Design

PascalPickel:
Added the new dialog design and some generic dialog creators for reusability

image


src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+            <Dialog.Header title="Test Title" />
+            <Dialog.Body>Test Text</Dialog.Body>
+            <Dialog.Footer>
+                <DialogButton onClick={noop}>Close</DialogButton>
+            </Dialog.Footer>
+        </Dialog>
+    );
+
+    describe('is visible when it', () => {
+        it('is rendered when visible', () => {
```
aadnekar, line 32: 'Dialog is visible when it is rendered when visible' ?

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+            render(dialog());
+            expect(screen.getByRole('dialog')).toBeInTheDocument();
+            test;
+        });
+
+        it('is not rendered when invisible', () => {
```
aadnekar, line 38: 'Dialog is visible when it is not rendered when invisible'

=> I suggest, remove the middle describe, rewrite the descriptions
e.g. ' Dialog is immediately visible by default' (The previous it())
and 'Dialog is hidden when you provide argument visible=true'

Please, if you don't mind, replace `it` with `test`. It's not important, but we have discussed earlier, and `test`, at least for me, makes more sense then the old `it` token.

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+            expect(screen.getByText('Test Text')).toBeInTheDocument();
+        });
+
+        it('in the footer', () => {
+            render(dialog());
+
+            expect(screen.getByText('Close')).toBeInTheDocument();
+        });
+    });
+});
```
aadnekar, line 63: These can all be a single test. Suggest to remove the describe and create a single test test('contains the expected content', () => {});

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+            expect(screen.getByText('Test Body')).toBeInTheDocument();
+        });
+
+        it('in the footer', () => {
+            render(dialog());
+
+            expect(screen.getByText('Close')).toBeInTheDocument();
+        });
+    });
```
aadnekar, line 90: Same as above, could be a single test.

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+    it('invokes the expected action', () => {
+        render(dialog());
+
+        const closeButton = screen.getByText('Close');
+        fireEvent.click(closeButton);
+
+        expect(noop).toHaveBeenCalled();
+    });
+});
```
aadnekar, line 137: Could even consider to just test all this in a single test, would then do it with the above as well. Then you don't need a describe and could just create a test in the body of the file.

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+        render(dialog());
+
+        fireEvent.click(screen.getByText('Optional'));
+        expect(noop).toHaveBeenCalled();
+        fireEvent.click(screen.getByText('ConfButton'));
+        expect(noop).toHaveBeenCalledWith(true);
+        fireEvent.click(screen.getByText('Cancel'));
+        expect(noop).toHaveBeenCalledWith(false);
+    });
+});
```
aadnekar, line 184: Same as above

src/Dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
...
+    .modal-content {
+        border: 0 transparent;
+        box-shadow: 0px 3px 6px #00000029;
+        background: #ffffff 0% 0% no-repeat padding-box;
+    }
+
+    .modal-header {
+        height: 40px;
+        border: 0 transparent;
+        background: $gray-50;
```
aadnekar, line 19: Could just let this be a `.css` file. To use a variable you'd just use the syntax:

```css
background: var(--gray-50);
```

src/Dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-4-Clause
+ */
+
+@import '../variables';
```
aadnekar, line 7: Suggestion: create a regular `.css` file instead. Don't remember exactly why, but there was a proposition to use regular `.css` files whenever we could.

You don't even need the import in a regular `.css` file, believe you'll just find the variables runtime.

src/ErrorDialog/ErrorDialog.tsx
```diff
@@ -25,40 +24,42 @@ const ErrorDialog = () => {
     const isVisible = useSelector(isVisibleSelector);
     const messages = useSelector(messagesSelector);
 
-    const defaultErrorResolutions = { Close: () => dispatch(hideDialog()) };
+    const defaultErrorResolutions = {
+        Close: () => dispatch(hideDialog()),
```
aadnekar, line 28: Do we not use *camelCase* for functions, or is this a special case where we don't? 
`Close --> close`

PascalPickel: The `Close` here is being used as a label for the button text.

src/Dialog/Dialog.test.tsx
PascalPickel: Fixed

src/Dialog/Dialog.test.tsx
PascalPickel: Fixed

src/Dialog/Dialog.test.tsx
PascalPickel: Fixed

src/Dialog/Dialog.test.tsx
PascalPickel: Fixed

src/Dialog/Dialog.test.tsx
PascalPickel: I combined the `expected content` ones but like the action one being seperate for now.
Will change this if you really want me to.

src/Dialog/dialog.scss
PascalPickel: I'd prefer to not mix css and scss files in the projects, though I don't see a major difference in this case in general.
I'll keep it as scss for now.

src/Device/DeviceSetup/DeviceSetupView.tsx
```diff
@@ -51,39 +51,60 @@ export default class DeviceSetupDialog extends React.Component<Props, State> {
...
-                                name="radioGroup"
-                                type="radio"
-                                disabled={isInProgress}
-                                onClick={() => this.onSelectChoice(choice)}
-                                label={choice}
-                            />
-                        ))}
-                    </Form.Group>
-                </ConfirmationDialog>
+                <Dialog isVisible={isVisible}>
```
jonasem, line 54: If this is a confirmation dialog, why not use the convenience creator function?

src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,167 @@
...
+    children: ReactNode | string;
+    confirmLabel?: string;
+    onConfirm: () => void;
+    cancelLabel?: string;
+    onCancel: () => void;
+    optionalLabel?: string;
+    onOptional?: () => void;
+}
+
+export const ConfirmationDialog = ({
```
jonasem, line 142: Export from shared as well

PR #459, opened 2023-01-04
Feat/persist serialport

aadnekar:
### Motivation

To allow different application to auto connect a device after a power cycle, we would like to persist the settings.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,6 +5,8 @@
  */
 
 declare module 'pc-nrfconnect-shared' {
+    import type { AutoDetectTypes } from '@serialport/bindings-cpp';
+    import { SerialPortOpenOptions } from 'serialport';
     import { Reducer, AnyAction } from 'redux';
     import React from 'react';
     import { Logger, LogEntry } from 'winston';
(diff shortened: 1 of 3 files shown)


src/Device/deviceLister.ts
```diff
@@ -43,42 +41,84 @@ export const wrapDeviceFromNrfdl = (device: NrfdlDevice): Device => ({
 export const wrapDevicesFromNrfdl = (devices: NrfdlDevice[]): Device[] =>
     devices.map(wrapDeviceFromNrfdl);
 
-/**
+const filterDeviceTraits = (lhs: DeviceTraits, rhs: DeviceTraits) => {
```
aadnekar, line 44: Could you please rewrite this function, in order to provide a bit more understanding to the reader.

src/Device/deviceLister.ts
```diff
@@ -43,42 +41,84 @@ export const wrapDeviceFromNrfdl = (device: NrfdlDevice): Device => ({
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        if (!(rhs as any)[key]) return true;
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        result = (lhs as any)[key];
+        return !result;
+    });
+
+    return !result;
+};
```
aadnekar, line 57:
```suggestion
// Returns false if device is to be included, true if device is to be filtered.
const filterDeviceTraits = (
    actualTraits: DeviceTraits,
    testTraits: DeviceTraits
) => {
    let result = false;

    Object.keys(testTraits).every(rule => {
        // If rule is false, skip the rule.
        if (!(testTraits as Record<string, boolean>)[rule]) return true;

        // If rule is true, check if deviceTrait is true.
        result = (actualTraits as Record<string, boolean>)[rule];
        return !result;
    });

    return !result;
};
```

aadnekar (commented):
I've got a few comments.

Look at the filter function especially, maybe we should discuss it :)

PR #471, opened 2023-01-18
Fix: Broken usage of unknown state

datenreisender:
The components `ErrorBoundary` and `Logo` were broken when used in the previous way from the launcher.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,8 +10,8 @@ and this project adheres to
 
 ### Fixed
 
--   The ErrorBoundary component was broken if the Redux state wasn't in a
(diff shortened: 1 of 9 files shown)


PascalPickel (approved):
LGTM

Do you expect more changes to shared? 
Otherwise you could just make this version 6.17.1 right?

datenreisender:
> Do you expect more changes to shared? Otherwise you could just make this version 6.17.1 right?

I first wanted you to have a look at it and test it myself in the launcher. That also looks good, so I will now do everything to release it as 6.17.1. :-)

PR #484, opened 2023-02-02
Improve persist store data

kylebonnici:


kylebonnici:
Persist serial port settings but use the app name from package to reduce possible issues 
Terminal Settings vCom indexes are now appended with vCom- to be more explicit in the saved file.

PR #485, opened 2023-02-02
Always regenerate types in full

kylebonnici:


datenreisender:
I assume this makes the generation slower. So why is this done? Did the incremental generation lead to bugs?

PR #510, opened 2023-02-24
Update to npm 18

kylebonnici:


kylebonnici:
We should make sure all apps and launcher can work with the new NPM and node first right?

PR #512, opened 2023-02-27
fix: properties not on global Window

aadnekar:
Explicitly add data directories to the global Window interface, in order to make typescript aware that the attributes should exist.

### Problem

These are the errors after running `npm run check` in pc-nrfconnect-cellularmonitor

image

### Solution

Append the relevant attributes to the global `Window` interface.

### DIFF
--- a/src/utils/appDirs.ts
+++ b/src/utils/appDirs.ts
@@ -11,6 +11,14 @@ import { loadPackageJson } from './packageJson';
 
 const getUserDataDir = () => getGlobal('userDataDir');
 
+declare global {
+    interface Window {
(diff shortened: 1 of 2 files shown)


kylebonnici (approved):

aadnekar:
> 

I was not aware of this, but I guess it's not picked up since the interface here is not in the global namespace?
That's why we added the `declare global`, so that the interface will be added to the `Window` interface found in the global namespace.
https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html

PR #520, opened 2023-03-08
fix concurrent port opening

kylebonnici:


src/SerialPort/SerialPort.ts
```diff
@@ -86,11 +89,31 @@ export const createSerialPort = async (
...
+                SERIALPORT_CHANNEL.OPEN,
+                options,
+                overwriteOptions
+            )) as string;
+        } catch (error) {
+            if (
+                (error as Error).message.includes(
+                    'PORT_IS_ALREADY_BEING_OPENED'
+                ) ||
+                retryCount === 0
```
aadnekar, line 104: Is this the intended logic? Should it not check if it has retries left?

```suggestion
                retryCount > 0
```

src/SerialPort/SerialPort.ts
```diff
@@ -86,11 +89,31 @@ export const createSerialPort = async (
         ipcRenderer.invoke(SERIALPORT_CHANNEL.SET, path, newOptions);
     };
 
-    const error = await ipcRenderer.invoke(
-        SERIALPORT_CHANNEL.OPEN,
-        options,
-        overwriteOptions
-    );
+    const tryAndOpen = async (retryCount: number) => {
```
aadnekar, line 92: This is a bit nitpicky, but could you rename the function? `tryAndOpen`, sounds like you try something and then you open. I suggest `tryOpen`, or `openWithRetries`.

src/SerialPort/SerialPort.ts
```diff
@@ -86,11 +89,31 @@ export const createSerialPort = async (
...
+            if (
+                (error as Error).message.includes(
+                    'PORT_IS_ALREADY_BEING_OPENED'
+                ) ||
+                retryCount === 0
+            ) {
+                return new Promise<string>(resolve => {
+                    setTimeout(async () => {
+                        resolve((await tryAndOpen(retryCount - 1)) as string);
+                    }, 50 + Math.random() * 100);
```
aadnekar, line 109: Does it need to be random? This will wait from `50ms` and up to `149ms`, can't it simply wait `100ms`?

aadnekar (commented):
The changes seems good to me, but could you check my comments before merging it?

PR #525, opened 2023-03-09
Fix/number inline input emit only on change

kylebonnici:


datenreisender:
Respecting a callback only on changed values sounds like an optimisation, that the users of this compoent should usually do themselves, when they need to. So, to me this doesn't sound like a fix but rather an optimsation. But if you prefer that we do it here, we can also do that.

datenreisender:
The logic for this is scattered around the component a bit. I wonder if it would be easier to understand if we encapsulate it in a hook on its own. E.g. something like a change signalling state which then could be used like 

```ts
  const [internalValue, setInternalValue, isChangedInternalValue] = useChangeSignallingState(externalValue));
```

But maybe that is too short-sighted by me and not really possible.

datenreisender:
Do I understand the code correctly, that this also changes the behaviour of arrow up and down that before the callback was always invoked, regardless of whether the new value was valid or not but now the callback is only invoked for valid values?

kylebonnici:
With regards to the keyboard arrow up and down:

- Old usage used to update the external value directly and hence would trigger `useSynchronisationIfChangedFromOutside` this unfortunately made it impossible for to detect is the value has changed or not. 
- Now the `onKeyboardIncrementAction` and `onKeyboardDecrementAction` pass in the new value which would be the same as if the user typed it. It is only a cosmetic change in my opinion one that was needed to achieve the goal of this PR

With regards to if this is a fix or an optimization I would say both: 

- it is an optimization because we don't emit unnecessary events
- fix because the wording of the events are `onChange` and  `onChangeComplete` and if no change occurred as a user of the component I would not expect to get the event. Also I don't see any reason a user of the component will need to get these events as if nothing changes no action is needed.

src/InlineInput/InlineInput.tsx
```diff
@@ -67,25 +67,28 @@ const InlineInput = React.forwardRef<HTMLInputElement, Props>(
...
 
-            const newValue = event.target.value;
             setInternalValue(newValue);
             if (isValid(newValue)) {
-                onChange(newValue);
+                if (beforeChangeValue !== newValue) {
+                    setBeforeChangeValue(newValue);
+                    onChange(newValue);
+                }
             }
```
aadnekar, line 92: I'm not able to immediately see the reason for having `beforeChangeValue`. Can you not just call `isValid(newValue)` before `setInternalValue`?

aadnekar (commented):
I would like to discuss the use case of the three different states that are used in the code, as I'm not able to see why they are needed. Other then that, I think that the code looks clean.

```javascript
const [internalValue, setInternalValue] = useState(externalValue);
const [beforeChangeValue, setBeforeChangeValue] =
            useState(externalValue);
const [initialValue, setInitialValue] = useState(externalValue);
```

datenreisender:
Regarding the rationale of the PR: I'd like to look at `onChange` and `onChangeComplete` separately: 

### `onChange`
I can see a motivation to _not_ call the callback if the new internal value is equal to the external value provided to the component. For this I think https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/727b0f1d0a8541059bc9f771bac80539c68c7dc4/src/InlineInput/InlineInput.tsx#L87 only needs to be changed to  
```ts
if (isValid(newValue) && newValue !== value) {
```

### `onChangeComplete`
This one is more tricky and depends on how you understand it. My intention of the function is close to `onBlur`, that it signifies that users finished editing a value. And, yes, simila to `onBlur` I think even if the final value is equal to the original value, then this should be called. So, the emphasis is less on the `change` and more on the `complete`.

Think about the situation, if users change a value from `12` to `123` and then back to `12` before confirming the value with enter or by leaving the input. Currently this would trigger three callbacks:
- `onChange(123)`
- `onChange(12)`
- `onChangeComplete(12)`

If the last one is not invoked, then it is impossible for the app to know whether the users stopped editing the value.

I understand that there are apps that have code that must only be called if the value was actually changed. But this is something that the app can detect themselves, by storing the original value and comparing to that, so I think this should be done by the app.

PR #526, opened 2023-03-13
Added feature for steppers

chunfantasy:
Added feature for steppers
image


src/Steppers/Steppers.tsx
```diff
@@ -0,0 +1,69 @@
...
+    warn?: boolean;
+    fail?: boolean;
+};
+
+export interface Props {
+    title?: string;
+    steps: Step[];
+}
+
+const Steppers: FC<Props> = ({ title, steps }) => (
```
aadnekar, line 27:
```suggestion
export interface Steppers {
    title?: string;
    steps: Step[];
}

const Steppers = ({ title, steps }: Steppers) => (
```

aadnekar: If I'm not mistaken, we usually just add the types of the props in the parameters. And typescript has a cool feature that lets the type definition live together with the value-definition. Meaning that both the `type Steppers` and the functional component `const Steppers` can live together.

So if you choose to use this approach, remember to remove the `{ FC }` dependency as well.

src/Steppers/Steppers.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <>
+        {title && <div>{title}</div>}
```
aadnekar, line 29: There are different preferences on this topic, but I prefer
```suggestion
        {title != null ? <div>{title}</div> : null}
```
And we could also check if the string is empty, because then it would not make sense to render the title, right?
```suggestion
        { (title != null && title !== '') ? <div>{title}</div> : null}
```

src/Steppers/Steppers.tsx
```diff
@@ -0,0 +1,69 @@
...
+        {steps.map(step => {
+            console.log(step);
```
aadnekar, line 31: Remove console.log

src/Steppers/Steppers.tsx
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2015 Nordic Semiconductor ASA
```
aadnekar, line 2:
```suggestion
 * Copyright (c) 2023 Nordic Semiconductor ASA
```

src/Steppers/steppers.scss
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright (c) 2015 Nordic Semiconductor ASA
```
aadnekar, line 2:
```suggestion
 * Copyright (c) 2023 Nordic Semiconductor ASA
```

aadnekar (approved):
Implementation looks good, there's nothing blocking this PR, in my opinion. But please review my comments, and suggested changes.

Generally, you could also consider to use more Bootstrap Utility Classes, https://getbootstrap.com/docs/4.3/layout/utilities-for-layout/, but as these class are well defined in the stylesheet, you could also argue that having them in the stylesheet makes more sense.

All in all, looks good! :rocket:

PR #530, opened 2023-03-14
Fix Error Boundary Button size

kylebonnici:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "pc-nrfconnect-shared",
-    "version": "19.0.0",
+    "version": "20.0.0",
```
KievDevel, line 3: 21 now I think :)
Also I'd suggest bumping shared version in a different PR

Changelog.md
```diff
@@ -9,6 +9,12 @@ every new version is a new major version.
 
 ## 19 - 2023-03-13
 
+### Fix
```
KievDevel, line 12: Would be nice to have main merged into current branch and updated accordingly (changes will now relate to v21)

PR #537, opened 2023-03-20
Feat/improve persisted serialport

PascalPickel:


src/Device/deviceSlice.ts
```diff
@@ -123,11 +127,50 @@ const slice = createSlice({
...
+            state,
+            action: PayloadAction<SerialPortOpenOptions<AutoDetectTypes>>
+        ) => {
+            if (state.selectedSerialNumber === null) return;
+
+            const vComIndex = state.devices
+                .get(state.selectedSerialNumber)
+                ?.serialPorts?.findIndex(e => e.path === action.payload.path);
+
+            if (vComIndex) {
```
kylebonnici, line 167: if vComIndex is 0 this will fail 

consider if (vComIndex !== undefined)

src/utils/persistentStore.ts
```diff
@@ -53,7 +53,7 @@ export const persistSerialPort = (
         lastUpdated: Date.now(),
         vComIndex,
     } as SerialSettings);
```
kylebonnici, line 55: Consider omiting the path from this type SerialSettings.serialPortOptions

kylebonnici (approved):

src/Device/deviceSlice.ts
aadnekar: Or `vComIndex >= 0` 🤷

PR #556, opened 2023-03-28
fix: handle missing nrf-udev installation crash

PascalPickel:


src/Device/deviceInfo/deviceInfo.ts
```diff
@@ -190,18 +191,32 @@ const deviceByUsb = (device: Device) => {
...
     }
     return null;
 };
 
-const unknownDevice = (device: Device): DeviceInfo => ({
-    name: device.usb?.product,
-    icon: isNordicDevice(device) ? unknownNordicLogo : unknownLogo,
-    website: {},
-});
+let hasShownUdevWarning = false;
```
kylebonnici, line 201: being pedantic:  hasShownUDevWarning  sry :P

kylebonnici (approved):

src/Device/deviceInfo/deviceInfo.ts
PascalPickel: I respectively decline

PR #575, opened 2023-04-12
[Work in progress] Switch to using nrf-fw-update-bundle-lib for dfu generation

michaeljones:
*This is a work in progress pull request aimed at sharing the potential changes for using the `nrf-fw-update-bundle-lib` to generate S/DFUs. Not looking to merge in this state.* 

This is an initial attempt at using the library instead of the current typescript-based dfu generation code.

These changes aim for minimal disruption and I haven't removed any dead code that is no longer used.

This change will break the pc-nrfconnect-programmer code base which relies on the exported createDfuZipBuffer function which we've changed here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,6 +74,7 @@
                 "lodash.range": "3.2.0",
                 "mousetrap": "1.6.5",
                 "npm-run-all2": "5.0.2",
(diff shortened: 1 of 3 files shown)


michaeljones:
I'm going to close this for the moment as my time is being redirected. I'll re-open it when I have more time to explore it.

PR #581, opened 2023-04-19
Update Dialog.tsx

kylebonnici:


aadnekar:
@kylebonnici What's the visual difference between using `<p>` and `<div>` in this context? Do you have a screenshot? What's the reason for this change?

kylebonnici:
This fixes the warning from react "Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>."

Visually it will look the same in this case

PR #590, opened 2023-04-28
Expose `getOptions` on SerialPort

aadnekar:
Typical use case for getOptions is for an app that is trying to connect to a port cannot because the port is claimed with different settings. Then, with getOptions, you'll be able to adapt to the current settings.

### DIFF
--- a/src/SerialPort/SerialPort.ts
+++ b/src/SerialPort/SerialPort.ts
@@ -179,3 +179,17 @@ export const createSerialPort = async (
         },
     };
 };
+
+export const getSerialPortOptions = async (
+    path: string
+): Promise<SerialPortOpenOptions<AutoDetectTypes> | undefined> => {
+    try {
+        console.log('will fetch options from path=', path);
+        return (await ipcRenderer.invoke(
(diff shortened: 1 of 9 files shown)


src/SerialPort/SerialPort.ts
```diff
@@ -89,6 +89,9 @@ export const createSerialPort = async (
         ipcRenderer.invoke(SERIALPORT_CHANNEL.SET, path, newOptions);
     };
 
+    const getOptions = (): Promise<SerialPortOpenOptions<AutoDetectTypes>> =>
```
datenreisender, line 92: In https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/blob/b4e8c2e4b0941cc7a838010b1bc9dabedff25974/src/ipc/serialport.ts#L55 it says  the return value is `SerialPortOpenOptions<AutoDetectTypes> | undefined`, here just `SerialPortOpenOptions<AutoDetectTypes>`.

I think the `| undefined` needs to be added here too.

aadnekar: Thanks, good catch :)

PR #606, opened 2023-05-11
Add Progress bar to device setup dialog

kylebonnici:


src/Device/DeviceSetup/DeviceSetupView.tsx
```diff
@@ -61,26 +73,43 @@ export default class DeviceSetupDialog extends React.Component<
...
+                            {progressMessage !== undefined && (
+                                <Form.Label>
+                                    <strong>Status:</strong>
+                                    <span>{` ${progressMessage}`}</span>
+                                </Form.Label>
+                            )}
+                            {progress !== undefined && (
+                                <ProgressBar
+                                    now={progress}
+                                    style={{ height: '4px' }}
```
PascalPickel, line 101: Extract this

src/Device/sdfuOperations.ts
```diff
@@ -111,6 +119,7 @@ const updateBootloader = async (
 
     logger.debug('Starting Bootloader Update');
 
+    dispatch(setDeviceSetupProgress(0));
```
PascalPickel, line 122: This should be added to the program function in jprogOperations also, right?

src/Device/sdfuOperations.ts
```diff
@@ -32,18 +36,22 @@ export type PromiseChoice = (
...
+            progress.progressPercentage
+        }%`;
+
+        if (status !== lastMSG) {
+            dispatch(setDeviceSetupProgress(progress.progressPercentage));
+            dispatch(setDeviceSetupProgressMessage(status));
+            logger.info(status);
+            lastMSG = status;
+        }
+    };
```
PascalPickel, line 54: Would like to see this extracted for reuse since it's the same as in jprogOperations.ts

PR #607, opened 2023-05-12
Refactor/device setup

kylebonnici:


typings/generated/src/Device/deviceSetup.d.ts
```diff
@@ -1,28 +1,50 @@
...
+    fw: string;
+    fwIdAddress: number;
+    fwVersion: string;
+}
+type PromiseChoice = (question: string, choices: string[]) => Promise<{
+    choice: string;
+    index: number;
+}>;
+export type PromiseConfirm = (message: string) => Promise<boolean>;
+export interface IDeviceSetup {
```
datenreisender, line 24: Rename `IDeviceSetup`.

We do not use `I` as a prefix for interfaces. Here it is even worse because there is `DeviceSetup` and `IDeviceSetup`.

datenreisender (changes requested):
This is probably still WIP, so consider this just to be a reminder that you mention the API changes in the changelog entry, e.g. I noticed that the shape of the `DeviceSetup` is changed.

src/Device/deviceSlice.ts
```diff
@@ -150,6 +150,7 @@ const slice = createSlice({
 
         setDeviceSetupProgress: (state, action: PayloadAction<number>) => {
             state.progress = action.payload;
+            state.isSetupDialogVisible = true;
```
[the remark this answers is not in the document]
kylebonnici: The case was that we might now need input (such as in the hex recovery) but we still want to show progress. I am trying to find a better way to do this :)

Changelog.md
```diff
@@ -7,16 +7,34 @@ This project does _not_ adhere to
...
 every new version is a new major version.
 
-## 46 - UNRELEASED
+## 46 - 2023-05-19
 
 ### Added
 
 -   Progress bar to device setup dialog
+-   Prepare device can be used from apps to be able program device at any point
+    such as Recover Hex for readback protected devices
```
PascalPickel, line 16:
```suggestion
-   `prepareDevice` function is now and can be used to program devices on demand
```

Changelog.md
```diff
@@ -7,16 +7,34 @@ This project does _not_ adhere to
...
+-   Custom devices setups can be now done by implementing IDeviceSetup and
+    passing it to `deviceSetup` array in `DeviceSetup`
```
PascalPickel, line 18:
```suggestion
-   Custom device setups can be now done by implementing `IDeviceSetup` and
    passing it as the `deviceSetup` array in `DeviceSetup`
```

typings/generated/src/Device/deviceSetup.d.ts
kylebonnici: Any suggestion on this @datenreisender not sure what to name it.

PR #608, opened 2023-05-19
feat: move needSerialport check into supportProgrammingMode

PascalPickel:
Additionally, the check for if the serialport is available has been removed. This is because this should be up to the app to handle (which is generally being done already).


kylebonnici (approved):
Looks good to me

PR #616, opened 2023-05-30
Dialogs now allow text to break and not overflow

kylebonnici:


datenreisender:
This change makes words break anywhere in dialogs and I think this is wrong. E.g. this element:

```JSX
<InfoDialog isVisible onHide={() => {}}>
    This is some text. This is some text. This is some text.
    This is some text. This is some text. This is some text.
    This is some text. This is some text. This is some text.
    This is some text. This is some text. This is some text.
</InfoDialog>
```

is rendered like this:
image

PR #629, opened 2023-06-09
Give more user friendly list of info collected

aadnekar:
The link to the nodejs documentation could perhaps confuse more then it was helping the user understand we are collecting the operating system of the OS the app is built for, and not what the user is running.


datenreisender (approved):
👍

PR #636, opened 2023-06-20
feat: Flash Messages

aadnekar:
In order to use flash messages, add the <FlashMessages /> component in the app. In order to emit a flash message call one of the relevant thunk functions exported from `flashMessageSlice.ts`.

### Added dependencies

For MDI icons convenience.

-  "@mdi/js": "^7.2.96"
- "@mdi/react": "^1.6.1"

### Flash Messages without `dismissTime`

image

### Flash messages with `dismissTime`

image

### Copied to clipboard

image

### DIFF
--- a/src/FlashMessage/FlashMessage.tsx
+++ b/src/FlashMessage/FlashMessage.tsx
@@ -33,7 +33,7 @@ const FlashMessage = ({ flashMessage }: FlashMessageProps) => {
     const [fadeoutTimer, setFadeoutTimer] = useState<string>(
         dismissTime == null ? 'unset' : `${dismissTime}ms`
     );
-    const timeoutHandler = useRef<NodeJS.Timeout | null>(null);
+    const timeoutHandler = useRef<NodeJS.Timeout | undefined>(undefined);
 
     if (timeoutHandler.current == null && dismissTime != null) {
         timeoutHandler.current = setTimeout(() => {
@@ -42,26 +42,22 @@ const FlashMessage = ({ flashMessage }: FlashMessageProps) => {
     }
 
     const close = () => {
-        if (timeoutHandler.current) {
-            clearTimeout(timeoutHandler.current);
-        }
+        clearTimeout(timeoutHandler.current);
         dispatch(removeMessage(id));
     };
 
     const addFadeout = () => {
-        if (dismissTime != null) {
+        if (dismissTime) {
             timeoutHandler.current = setTimeout(() => {
                 dispatch(removeMessage(id));
             }, dismissTime);
+            setFadeoutTimer(`${dismissTime}ms`);
         }
-        setFadeoutTimer(`${dismissTime}ms`);
     };
 
     const removeFadeout = () => {
+        clearTimeout(timeoutHandler.current);
         setFadeoutTimer('unset');
-        if (timeoutHandler.current) {
-            clearTimeout(timeoutHandler.current);
-        }
     };
 
     const initialRender = () => divRef.current == null;
@@ -80,7 +76,6 @@ const FlashMessage = ({ flashMessage }: FlashMessageProps) => {
                 flexDirection: 'column',
                 justifyContent: 'space-between',
             }}
-            className={`flash-variant-${variant}`}
             onMouseEnter={removeFadeout}
             onMouseLeave={addFadeout}
         >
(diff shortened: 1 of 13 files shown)


aadnekar:
Questions:

1. Is it possible to automatically add the `messages: flashMessagesReducer` in the app specific redux store, instead of having to add it in each app separately.
2. Should we have an upper limit to the number of Flash Messages? Currently set to ` MAX_NUMBER_OF_MESSAGES = 8`: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/889bae8c6d97567ff28fffeea412aed7b46cc02a/src/FlashMessage/FlashMessageSlice.ts#L17

jonasem:
> Questions:
> 
> 1. Is it possible to automatically add the `messages: flashMessagesReducer` in the app specific redux store, instead of having to add it in each app separately.
> 2. Should we have an upper limit to the number of Flash Messages? Currently set to ` MAX_NUMBER_OF_MESSAGES = 8`: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/889bae8c6d97567ff28fffeea412aed7b46cc02a/src/FlashMessage/FlashMessageSlice.ts#L17

1. I think you want to add it to the rootReducer along with the other stuff from shared that's in redux.
2. Not sure what should happen if someone pushes more than the max then, should they be queued until the visible ones are destroyed?

src/FlashMessage/FlashMessage.tsx
```diff
@@ -0,0 +1,160 @@
...
+    return (
+        <div
+            style={{
+                width: '256px',
+                gap: '16px',
+                marginBottom: '32px',
+                display: 'flex',
+                flexDirection: 'column-reverse',
+            }}
+            className="message-container"
```
jonasem, line 133: Is this a bootstrap class?

src/FlashMessage/FlashMessage.tsx
```diff
@@ -0,0 +1,160 @@
...
+            return colors.red;
+        case 'success':
+            return colors.green;
+        case 'info':
+            return colors.nordicBlue;
+        case 'warning':
+            return colors.orange;
+
+        default:
+            return '' as never;
```
jonasem, line 157: Consider using `'unset'` as a fallback css value

src/FlashMessage/FlashMessage.tsx
```diff
@@ -0,0 +1,160 @@
...
+                color: colors.white,
+                zIndex: 1000,
+                animation: initialRender() ? 'slide-in 1s' : 'unset',
+                width: '100%',
+                padding: '16px',
+                display: 'flex',
+                flexDirection: 'column',
+                justifyContent: 'space-between',
+            }}
+            className={`flash-variant-${variant}`}
```
jonasem, line 83: Seems like an unused class?

src/FlashMessage/FlashMessage.tsx
aadnekar: I agree, but I would like it to return `as never`, to indicate to the programmer that we only accept the explicitly handled variants.

aadnekar:
> > Questions:
> > 
> > 1. Is it possible to automatically add the `messages: flashMessagesReducer` in the app specific redux store, instead of having to add it in each app separately.
> > 2. Should we have an upper limit to the number of Flash Messages? Currently set to ` MAX_NUMBER_OF_MESSAGES = 8`: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/889bae8c6d97567ff28fffeea412aed7b46cc02a/src/FlashMessage/FlashMessageSlice.ts#L17
> 
> 1. I think you want to add it to the rootReducer along with the other stuff from shared that's in redux.
> 2. Not sure what should happen if someone pushes more than the max then, should they be queued until the visible ones are destroyed?

1. If it's added to the rootReducer, then they the Flash Messages will be shared across apps, which is not what we want (I believe?).
2. Hmm, could be added to a queue, because if we do not have an upper limit some of the messages may be hidden anyway, because they are pushed outside of the browser window.

aadnekar:
> > 1. If it's added to the rootReducer, then they the Flash Messages will be shared across apps, which is not what we want (I believe?).
> 
> No. You can safely add it to the rootReducer because every app has its own store. Otherwise error dialogs or the whole log would also appear from one app in the other.

Very nice, just tested it :smile: Thanks for letting me know.

datenreisender:
I am curious: Why did you replace the counter to generate IDs with `nanoid`?

aadnekar:
> I am curious: Why did you replace the counter to generate IDs with `nanoid`?

Should have added a more verbose commit message. So, I was shown how to Redux Toolkit use the `prepare` property, and we saw that it also exposes the `nanoid` function. I was just curious about it, and it looks quite nice, so I wanted to implement it :smile:

jonasem (approved):

PR #638, opened 2023-06-20
Remove device if clearWaitForDevice called while timeout is already r…

kylebonnici:
…unning


datenreisender (approved):
Looks good.

Some changes which I think would make it better are in #643. Feel free to apply it, reject it, or ask me about it.

When you moved the function `removeDeviceFromList` out of `startWatchingDevices` to a top level function, I assume you didn't alter any behaviour but just adjusted the function to its new surrounding. When doing changes like this, they are easier for me to review and understand, when you do them in a commit on their own. Especially separating commits which contain refactorings from ones that intent to alter the behaviour (like wrapping `clearWaitForDevice` so that it now calls the `onCancelTimeout` callback) and mentioning in the commit message if something is supposed to change behaviour or not.

PR #640, opened 2023-06-21
Remove mousetrap

jonasem:


aadnekar:
Are there any blockers keeping this from being merged?

jonasem:
> Are there any blockers keeping this from being merged?

Not really, just want a review

PR #641, opened 2023-06-21
PR suggestions

jonasem:


src/FlashMessage/FlashMessage.tsx
```diff
@@ -42,55 +42,43 @@ const FlashMessage = ({ flashMessage }: FlashMessageProps) => {
...
-
     return (
         <div
             ref={divRef}
             style={{
                 backgroundColor: getBackgroundColorFromVariant(variant),
                 color: colors.white,
                 zIndex: 1000,
-                animation: initialRender() ? 'slide-in 1s' : 'unset',
+                animation: !divRef.current ? 'slide-in 1s' : 'unset',
```
aadnekar, line 71: I think the function name alone is a good enough reason to keep the function, as it gives some explanation to why you want `divRef.current` to be `null` or `undefined` in order to add the animation. This may not be trivial coming back to the code in 1 week time.

src/FlashMessage/FlashMessage.tsx
```diff
@@ -42,55 +42,43 @@ const FlashMessage = ({ flashMessage }: FlashMessageProps) => {
...
             onMouseLeave={addFadeout}
         >
-            <div
-                style={{
-                    width: '100%',
-                    display: 'flex',
-                    justifyContent: 'space-between',
-                }}
-            >
+            <div className="d-flex w-100 justify-space-between">
```
aadnekar, line 81: I thought we just agreed to not add more bootstrap utility classes, as we will port to tailwind?

src/FlashMessage/FlashMessage.tsx
```diff
@@ -154,7 +139,7 @@ const getBackgroundColorFromVariant = (variant: FlashMessageVariant) => {
             return colors.orange;
 
         default:
-            return '' as never;
+            return 'unset';
```
aadnekar, line 142: This would then indicate that a variant not explicitly handled in the switch statement is Ok, I would rather indicate that the default case should never happen.

```suggestion
            return 'unset' as never;
```

aadnekar (commented):
I like most of the suggestions, and think they improve my PR, but I've made a few comments that I would like to discuss.

aadnekar:
Will close, have added some of the suggestions in the original PR, in commit: 8b4538fbe17a61ae7f7f91d8b5001ff7c039e8de

PR #657, opened 2023-06-30
Export types AppThunk and AppDispatch globally

kylebonnici:


jonasem (approved):

src/store.ts
```diff
@@ -50,6 +50,8 @@ const store = (appReducer?: Reducer) =>
 
 // Needed only to infer the types below
 const concreteStore = store();
+
+export type TAction<T> = ThunkAction<T, RootState, undefined, AnyAction>;
```
jonasem, line 54: Is it possible to make `TAction` be an alias for `TAction<void>`?

datenreisender: In the Redux docs they are using `void` as the default for the return type. And they are calling it `AppThunk` instead of `TAction`. I suggest that we adopt both: 
```suggestion
export type AppThunk<ReturnType = void> = ThunkAction<
  ReturnType,
  RootState,
  unknown,
 AnyAction
>
```

https://redux.js.org/usage/usage-with-typescript#type-checking-redux-thunks

In the Redux docs they are also using `UnknownAction` instead of `AnyAction` but AFAICS we cannot use this yet, because it will be only part of the upcoming Redux 5.

PR #682, opened 2023-07-06
Feat/use project tailwind config

PascalPickel:


scripts/esbuild-renderer.js
```diff
@@ -74,9 +74,7 @@ function options(additionalOptions) {
             postCssPlugin({
                 postcss: {
                     plugins: [
-                        tailwindcss(
-                            './node_modules/pc-nrfconnect-shared/config/tailwind.config.js'
-                        ),
+                        tailwindcss('./tailwind.config.js'),
```
KievDevel, line 77: Can it be done in a way, that it will check if './tailwind.config.js' exists and if no - stick to default from shared?

KievDevel: Otherwise we'll have to update every project and include tailwind.config.js there

PascalPickel: We should update every project to have a config

PascalPickel: But I'll add a check for now. I'm against it but the option can always be removed later on

PR #683, opened 2023-07-06
Custom preflight class for tailwind use

KievDevel:


src/App/preflight.scss
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    /* Make elements with the HTML hidden attribute stay hidden by default */
+    [hidden] {
+        display: none;
+    }
+
+    p {
+        font-size: 20px;
+    }
```
PascalPickel, line 390: I cannot find this in tailwind base styles?
Where is this coming from?

src/App/preflight.scss
```diff
@@ -0,0 +1,391 @@
...
+    ul,
+    menu {
+        list-style: none;
+        margin: 0;
+        padding: 0;
+    }
+
+    dialog {
+        padding: 0;
+    }
```
PascalPickel, line 318: I cannot find this in tailwind base styles?
Where is this coming from?

src/App/preflight.scss
```diff
@@ -0,0 +1,391 @@
...
+    */
+
+    button,
+    input,
+    optgroup,
+    select,
+    textarea {
+        font-family: inherit; /* 1 */
+        font-feature-settings: inherit; /* 1 */
+        font-variation-settings: inherit; /* 1 */
```
PascalPickel, line 178: I cannot find this in tailwind base styles?
Where is this coming from?

src/App/preflight.scss
```diff
@@ -0,0 +1,391 @@
...
+    /*
+    Add the correct display in Chrome and Safari.
+    */
+
+    summary {
+        display: list-item;
+    }
+
+    /*
+    Removes the default spacing for appropriate elements.
```
PascalPickel, line 280:
```suggestion
    Removes the default spacing and border for appropriate elements.
```
This was the original comment I found in the base style linked by tailwind.

PascalPickel (commented):
Found some differences based on the the base style linked by the tailwind docs:
https://unpkg.com/tailwindcss@3.3.2/src/css/preflight.css

src/App/preflight.scss
KievDevel: Thanks for noticing, this one comes from testing, thought I did remove it.

src/App/preflight.scss
KievDevel: This comes from here https://github.com/tailwindlabs/tailwindcss/blob/2bfdbce964f36b57fbac1725c807805f1d7a6ad7/src/css/preflight.css#L269
...and I do not see any border styles below

PascalPickel: Ah, I was looking at the preflight linked to from the documentation. Not sure why they are different or if it matters which one we use?

KievDevel: I believe github repo version is just up to date. Shouldn't be a big diff for us.

PascalPickel (approved):

PR #694, opened 2023-07-18
Feat/nrfutil device

boundlesscalm:


boundlesscalm:
I'd like it if shared defined it's supported nrfutil versions and can tell us to update the versions in the app package.json.
Something like in the check-app-properties script

src/Nrfutil/sandbox.ts
```diff
@@ -0,0 +1,446 @@
...
+        }
+    ): Buffer | undefined => {
+        const parsedData: NrfutilJson<Result>[] | undefined =
+            parseJsonBuffers(data);
+
+        if (!parsedData) {
+            return data;
+        }
+
+        parsedData.forEach(item => {
```
boundlesscalm, line 110: All the magic casts for item.data should be handled by the types as proposed in the NrfutilJson comment

boundlesscalm:
Separate callback for when enumeration is closed.
- @kylebonnici

PR #700, opened 2023-07-21
Feat/tailwind button

kylebonnici:


src/Button/Button.tsx
```diff
@@ -39,21 +38,39 @@ const Button: React.FC<ButtonProps> = ({
...
+                variant === 'warning' &&
+                    'tw-bg-orange tw-text-white active:enabled:tw-bg-orange-700',
+                variant === 'danger' &&
+                    'tw-bg-red tw-text-white active:enabled:tw-bg-red-700',
+                variant === 'link' &&
+                    'tw-bg-transparent tw-p-0 tw-text-nordicBlue',
+                variant === 'link-button' &&
+                    'tw-border tw-border-nordicBlue tw-bg-white tw-text-nordicBlue active:enabled:tw-bg-gray-50',
+                className
+            )}`}
```
KievDevel, line 66: Would be nice to have this logic extracted somewhere else to improve readability. It can be just a configuration object instead.

src/Dialog/Dialog.tsx
```diff
@@ -78,7 +78,7 @@ Dialog.Body = ({ children }: { children: ReactNode }) => (
 );
 
 Dialog.Footer = ({ children }: { children: ReactNode }) => (
-    <Modal.Footer>{children}</Modal.Footer>
+    <Modal.Footer className="tw-flex tw-gap-1">{children}</Modal.Footer>
```
KievDevel, line 81: might be a breaking change. not sure, but doesn't look safe

src/Button/Button.tsx
kylebonnici: If I take them out we loose the lint ordering and language server information at least that is my experience here why these are here

KievDevel: Do you mean smth like this?
https://stackoverflow.com/questions/66614875/how-can-i-enable-tailwind-intellisense-outside-of-classname

kylebonnici: Problem is two fold:
1) if not associated with a var class/className we do not get language assistance like showing auto complete, showing the meaning of the class e.g. hovering on the class I get 

`.tw-text-white {
    --tw-text-opacity: 1 !important;
    color: rgb(255 255 255 / var(--tw-text-opacity)) !important;
}`

this is lost of not in a variable names class or className

We can add more regex but in the plugin settings but that is a developer thing and we cannot expect all of us to add multiple variables apart from the default as it will not be maintainable IMO

2) Even more problematic 

Formatting the order of the classes does not work if classnames are not a part of a an element inside ClassName property so solution for problem 1 will sill be a problem for this.

This helps sort all classes in the same order in out code and we cannot have this if classes are outside in some variable

PR #710, opened 2023-08-03
Feat/nrfutil sandbox impl

kylebonnici:


datenreisender:
Still looking at this (as good as it is possible without a normal working internet connection :-/), just two first comments:
- In `package.json` the `nrfutil` is added at the top level. Please let us instead add a `nrfConnectForDesktop` property at the top level and add `nrfutil` as a child to that property.
- I am surprised that this PR doesn't add any exports e.g. to `src/index.ts` or `main/index.ts`. Shouldn't something like `prepareSandbox` or `prepareAndCreate` be exported there, so that in the future the launcher can use this code to prepare the sandbox for an app?

src/Nrfutil/sandbox.ts
```diff
@@ -0,0 +1,420 @@
...
+                            os.platform() === 'win32' ? '.exe' : ''
+                        }`
+                    )
+                )
+            ) {
+                getModuleVersion()
+                    .then(moduleVersion =>
+                        resolve(moduleVersion.version === version)
+                    )
+                    .then(reject);
```
datenreisender, line 150: This probably was supposed to be `.catch(reject)`, right? The way it is currently, `resolve` and `reject` are always both called when the Promise returned by `getModuleVersion` resolves.

datenreisender:
- Do not split of the types, rather split by domain
	- E.g. move all the types regarding Module versions from sandboxTypes into `resolveModuleVersion.ts` (which is probably better named just `moduleVersion.ts`). 
	- In `device.ts` it could be good to extract each command into it's own file or at least do this by groups (e.g. listing devices, querying devices, programming), and always keep types and corresponding code together.
- The modules `export` many things where I am not sure why they are exported because they not used anywhere else. Is this by intention because they are designed to be used in other places or is this by oversight because e.g. you seemingly needed them somewhere else during development but then forgot to remove the `export`.

PR #715, opened 2023-08-08
Allow ThunkActions in the initial store prepared by render()

aadnekar:
In order to allow ThunkActions, which we commonly use in our apps, we accept both Action and ThunkAction in the actions array provided in the render() function.

#### Typical usecase

Extract from Power Profiler

```typescript
const initialStateActions = [
    chartWindowAction(1, 1_000_000, 1_000_000),
    showExportDialog(), // ThunkAction
];

test('someTest', () => {
        ...
        render(<ExportDialog />, initialStateActions); // Would error before, because it did not accept ThunkAction.
        ....
});
```


datenreisender:
Hi @aadnekar. 

I took the liberty of pushing my suggestions as new commits here. Feel free to remove them again if they do not work for you (or tell me, and I will remove them). 

As mentioned in the commit, it may require that you also use `AppThunk` in the app but I tried that in PPK and it seems to work.

PR #717, opened 2023-08-10
Enhance pipeline release to npm

datenreisender:
Some smaller enhancements.

The changes to the postinstall script are a bit obscured, because I also converted it to TypeScript during the branch. Besides the conversion I also did these three changes to it:
- Generate types only if needed
- Only compile the `nordic-publish` script
- Add logging for each step, so one can see what is done


jonasem (approved):
Tested ok on my windows machine as well 👍

PR #727, opened 2023-08-15
Nrfutil refactoring proposal

KievDevel:


nrfutil/device/batch.ts
```diff
@@ -133,21 +154,34 @@ export class Batch {
         count: number,
         callback: (completedTasks: TaskEnd<unknown>[]) => void
     ) {
+        if (!callback || typeof count === 'undefined') {
+            // handle if not passed: log use case or anything else
+            // same check can be applied to the methods above if it makes sense
+
+            return this;
+        }
```
datenreisender, line 162: Why add these checks when they should never happen according to the types? If we want to become more liberal on what we accept as arguments then we should also change the types of the parameters but I currently do not see an advantage in it.

nrfutil/device/batch.ts
```diff
@@ -133,21 +154,34 @@ export class Batch {
...
             operationId: this.operations.length - 1,
             count,
         });
 
         return this;
     }
 
-    public async run(
+    // here would be nice to have an option to pass timeout (like 2/5/10min etc) and allow user to abort operation if stuck
+    // otherwise we might be stuck in infinite await, right?
```
datenreisender, line 177: _If_ we need a timeout, it is already possible to implement it in the client of the function. They can abort the task at any time. If we need this very often, then we can add it later here.

nrfutil/device/list.ts
```diff
@@ -46,24 +46,36 @@ export default async (
     }
 
     const onData = (data: HotplugEvent | ListEvent) => {
+        // general check I write always. Keep if it makes sense.
+        // Safeguard if undefined is passed so accessing any object property won't break with exception
+        // same can be applied to the functions above
+        if (!data) return;
```
datenreisender, line 52: I generally recommend not to do that everywhere but only on the border of your system and on few important parts within. 

And then usually make it clear through the argument types that it handles undefined. 

And usually check explicitly for `null`/`undefined`, so e.g. falsy values like empty strings do not lead to surprising behaviour.

So, in short: Do not add a `if (!data) return;` here or in most other places.

nrfutil/device/list.ts
```diff
@@ -46,24 +46,36 @@ export default async (
...
-            } else if (data.event === 'Left') {
-                onHotplugEvent.onDeviceLeft(data.id);
-            }
+
+            return;
+        }
+
+        if (!onHotplugEvent || !isHotplugEvent(data)) {
+            return;
+        }
```
datenreisender, line 66: Doing a quick exit here, makes this more easy to understand. 👍

nrfutil/device/list.ts
```diff
@@ -46,24 +46,36 @@ export default async (
...
+
+        const newDeviceArrived =
+            data.event === 'Arrived' && data.device && data.device.serialNumber;
```
datenreisender, line 69: Can be simplified with optional chaining:
```suggestion
            data.event === 'Arrived' && data.device?.serialNumber;
```

nrfutil/sandbox.ts
```diff
@@ -26,10 +31,15 @@ import {
 } from './sandboxTypes';
 
 const parseJsonBuffers = <T>(data: Buffer): T[] | undefined => {
+    // bcz if data is undefined data.toString().trim() will throw an exception
+    if (!data) return undefined;
```
datenreisender, line 35: Like above, do not add this sanity check here. The clients of this function already take care of `data` always being defined.

nrfutil/device/batch.ts
KievDevel: That's an idea I've just shared here. Timeouts are also implemented inside device-lib, so we might not need that at all. I'm just not sure what logic we're losing by dropping device-lib-js.

nrfutil/device/list.ts
KievDevel: "only on the border of your system and on few important parts within" - definitely makes sense. This proposal has same motivation as one of the checks above ("Motivation behind...")

nrfutil/device/list.ts
KievDevel: That's the only purpose of this change. Thanks!

KievDevel:
Closed after review. Accepted changes & ideas were migrated into merged nrfutil branch by Kyle.

PR #728, opened 2023-08-15
Fix/html bundle

boundlesscalm:
Currently the official launcher doesn't support when an app is built with `cjs` and without externalised `react`.

I made the `html` property opt-in and also renamed it (as we want to have it in the `nrfConnectForDesktop` property.

The launcher pr using this change is: https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/pull/865

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,7 +7,7 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
-## 88 - 2023-08-15
+## 88 - 2023-08-17
 
 ### Changed
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,17 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 88 - 2023-08-15
+
+### Changed
+
+-   `html` package json property has been moved to `nrfConnectForDesktop.html`.
+-   `nrfConnectForDesktop.html` is now opt-in and apps can be built without it.
+
+### Steps to upgrade
+
+-   Change `html` to `nrfConnectForDesktop.html` in the app package json.
+
 ## 87 - 2023-08-14
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


jonasem (approved):

Changelog.md
```diff
@@ -7,6 +7,17 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 88 - 2023-08-15
+
+### Changed
+
+-   `html` package json property has been moved to `nrfConnectForDesktop.html`.
+-   `nrfConnectForDesktop.html` is not opt-in and apps can be built without it.
```
datenreisender, line 15: This doesn't seem to make sense to me. 

I think either should the “not” be “now” (`nrfConnectForDesktop.html` is now opt-in and apps can be built without it) or the “can” should be “cannot” (`nrfConnectForDesktop.html` is not opt-in and apps cannot be built without it.). I assume the first one is correct.

datenreisender (commented):
Maybe also add a comment in the changelog entry of v84 that doing the recommended changes there (adding the `html` property  and calling `render`) will make the app incompatible with the current launcher and thus the `engine` field should be bumped accordingly?

Changelog.md
boundlesscalm: Should be `now`. Thanks for catching that.

boundlesscalm:
> Maybe also add a comment in the changelog entry of v84 that doing the recommended changes there (adding the `html` property and calling `render`) will make the app incompatible with the current launcher and thus the `engine` field should be bumped accordingly?

I mentioned it in the "Steps to upgrade"

jonasem (approved):

Changelog.md
```diff
@@ -7,6 +7,17 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 88 - 2023-08-15
```
jonasem, line 10:
```suggestion
## 88 - 2023-08-17
```

boundlesscalm: Thanks for the suggestion. Fixes as per your orders : )

PR #733, opened 2023-08-22
Enhance nrfutil

datenreisender:
- e98f61c: Give the nrfutil module types more expressive names
- 58fef9b: Check the content of the `nrfutil` entry in `package.json`, e.g. make sure that there always is at least one entry. Being able to depend on this, makes e.g. code in the launcher simpler.
- dbcce77 + 5cebf2c: Remove the remains of `nrf-device-lib-js`.


kylebonnici (approved):
Nice :) 🤜🤛

PR #734, opened 2023-08-22
[FIX] Scrollbar overlays modal's content

KievDevel:
`overflow: overlay` is the reason why scrollbar overlays other content like modals.
Replacing with `overflow: scroll` is fixing this issue https://nordicsemi.atlassian.net/browse/NCD-157

Note: also affects any other areas where scrollbar is used. **If this solution is good to go, same should be applied to every area where scrollbar is used**

### DIFF
--- a/src/App/app.scss
+++ b/src/App/app.scss
@@ -67,7 +67,7 @@
 
     .core19-side-panel-container {
         width: $side-panel-width;
-        padding: 0 24px;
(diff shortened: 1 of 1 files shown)


kylebonnici (approved):

KievDevel:
Bug exists, but didn't find any use cases when content might become unavailable (log's scrollbars won't affect modals bcz of their position). Merging PR for now

PR #735, opened 2023-08-23
Feat/overlay

boundlesscalm:


aadnekar:
In what other app than Cellular Monitor do we use the Overlay? And as we have started moving away from Bootstrap, is it worth investigating to do this with an alternative to Bootstrap?
Might be worth mentioning that Bootstrap use Popper.js for its overlay/tooltip components: https://popper.js.org/

boundlesscalm:
> In what other app than Cellular Monitor do we use the Overlay? And as we have started moving away from Bootstrap, is it worth investigating to do this with an alternative to Bootstrap? Might be worth mentioning that Bootstrap use Popper.js for its overlay/tooltip components: https://popper.js.org/

NPM uses it and the idea is to use it in another app soon.
Sure it's worth investigating.
Putting it in shared just unifies the design, as well as make the move away from bootstrap easier.

kylebonnici (approved):

PR #758, opened 2023-09-01
Feat/nrfutil device uart write

kylebonnici:


Changelog.md
```diff
@@ -7,6 +7,12 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 106 - UNRELEASED
+
+### Added
+
+-   Nrfutil device `uart-write` support
```
aadnekar, line 14: Could you add some more description to what this feature does?
E.g. why do we want to add it, what is it's main purpose and are there other use cases?

aadnekar (approved):
It looks good to me, but it's my first step to looking into the nrfutil domain, so please do not only rely on my review.
I've added a small comment to the readme, which is not a must-do, but I would prefer more verbose descriptions in the readme, as this is my go-to place to look at changes in shared.

PR #765, opened 2023-09-06
Feat/feedback export

boundlesscalm:
Quickstart will be sending feedback but is not using the App component (which implicitly uses this feedbackpane). Additionally I can't use the feedback pane directly due to size, styling and logger imports.

Thus, I simply export a bare minimum `sendFeedback` function.


chunfantasy (approved):
LGTM

PR #783, opened 2023-09-28
Fix: keep showing content even if space is less then min width

kylebonnici:


src/MasonryLayout/MasonryLayout.tsx
```diff
@@ -193,12 +193,16 @@ export default ({
         };
 
         const action = () => {
-            const noOfColumns = Math.floor(
-                current.clientWidth /
-                    (minWidth + Number.parseInt(styles.margin, 10))
-            );
+            const noOfColumns =
```
aadnekar, line 196: Is this variable `numberOfColumns`, in that case, should it not be called `numberOfColumns`?

src/MasonryLayout/MasonryLayout.tsx
```diff
@@ -193,12 +193,16 @@ export default ({
...
+                current.clientWidth >= minWidth
+                    ? Math.floor(
+                          current.clientWidth /
+                              (minWidth + Number.parseInt(styles.margin, 10))
+                      )
+                    : 1;
```
aadnekar, line 202: Could you potentially extract this to a function?

```typescript
// Somewhere outside of the component:
const getNumberOfColumns = (minWidth: number, clientWidth: number) => 
    current.clientWidth >= minWidth ? Math.floor(...) : 1;
    
 // Inside the component ( where you now do the calculation):
 const numberOfColumns = getNumberOfColumns( ... );
```

src/MasonryLayout/MasonryLayout.tsx
```diff
@@ -236,7 +240,10 @@ export default ({
                     hiddenChildren={hiddenChildren}
                     width={masonryLayoutRef.current?.clientWidth ?? -1}
                     columns={columns}
-                    minWidth={minWidth}
+                    minWidth={Math.min(
+                        minWidth,
+                        masonryLayoutRef.current?.clientWidth ?? minWidth
+                    )}
```
aadnekar, line 246: This could potentially also be extracted to a function, in order to better communicate the intention of using Math.min(...). It's kind of difficult to keep track of what is the purpose of all the different variables.

E.g. `getMinimumWidthToContainContent()`

I may be wrong though 🤔

aadnekar (commented):
Was about to approve, but you could just skim through the comments, and see if you want to do any changes in the future 🤭

PR #789, opened 2023-10-11
Feat/nrfutil 2.0.0

kylebonnici:


boundlesscalm (approved):
Looks to be mostly just changes based on the changes to NrfUtilDevice.
Are there any other implications from this?
Otherwise it looks alright to me.
Just have to up the version and add a changelog entry.

PR #793, opened 2023-10-19
Add NumberInputWithDropdown component

aadnekar:
For now the implementation is quite specific for serial-terminal, but will hopefully evolve and be more generic in the future.

How it works is that when you click on the component, it will work as any other NumberInlineInput component, or if you click the chevron on the right, you will see a list of dropdown items to select from.

The implementation design-wise is equivalent to the Dropdown component.

**Baud Rate** is using the new NumberInputWithDropdown component, and **Data bits** is using the Dropdown component.
image

You can type in anything you want, but will see that the input is invalid when it is invalid.
image

When you click the Chevron, you see the list of items to select from. When an item is selected, the input field will immediately be updated to that value.
image

### DIFF
--- a/src/InlineInput/InlineInput.tsx
+++ b/src/InlineInput/InlineInput.tsx
@@ -58,6 +58,8 @@ interface Props {
     onKeyboardDecrementAction?: () => string;
     className?: string;
     textAlignLeft?: boolean;
+    onValidityChanged?: (validity: boolean) => void;
+    preventDefaultInvalidStyle?: boolean;
 }
 
 const InlineInput = React.forwardRef<HTMLInputElement, Props>(
@@ -72,19 +74,27 @@ const InlineInput = React.forwardRef<HTMLInputElement, Props>(
             onKeyboardDecrementAction = () => externalValue,
             className = '',
             textAlignLeft = false,
+            onValidityChanged,
+            preventDefaultInvalidStyle,
         },
         ref
     ) => {
         const [internalValue, setInternalValue] = useState(externalValue);
         const [initialValue, setInitialValue] = useState(externalValue);
         useSynchronisationIfChangedFromOutside(externalValue, setInternalValue);
+
         const onChangeIfValid = (newValue: string) => {
             if (disabled) {
                 return;
             }
 
+            const validity = isValid(newValue);
+            if (onValidityChanged != null) {
+                onValidityChanged(validity);
+            }
+
             setInternalValue(newValue);
-            if (isValid(newValue)) {
+            if (validity) {
                 if (externalValue !== newValue) {
                     onChange(newValue);
                 }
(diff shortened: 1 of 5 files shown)


src/InlineInput/NumberInputWithDropdown.tsx
```diff
@@ -0,0 +1,121 @@
...
+                if (!event.currentTarget.contains(event.relatedTarget)) {
+                    setIsActive(false);
+                }
+            }}
+        >
+            <FormLabel className="tw-mb-1 tw-text-xs">{label}</FormLabel>
+            <div className="tw-flex tw-h-8 tw-w-full">
+                <NumberInlineInput
+                    value={inlineValue}
+                    range={{ min: 1, max: 1_000_000, step: 1, decimals: 0 }}
```
aadnekar, line 66: This is wrong, should be a property to the component and then passed to the `NumberInlineInput`

aadnekar:
Changes:

Very minimalistic invalid state, but should be quite visible and intuitive
image
image

The "normal" state now has a white underline in order to communicate to the user that it is interactable
image
image

src/InlineInput/InlineInput.tsx
```diff
@@ -142,12 +144,12 @@ const InlineInput = React.forwardRef<HTMLInputElement, Props>(
...
                 ref={ref}
                 type="text"
-                className={classNames(
+                className={`${classNames(
                     'inline-input',
                     isValid(internalValue) || 'invalid',
                     disabled && 'disabled',
                     className
-                )}
+                )} ${textAlignLeft && 'tw-pl-2 tw-text-left'}`}
```
datenreisender, line 152: If `textAlignLeft` is `false`, the string `false` will be added to the class names.

Better use the `classNames` function, which is already there anyhow, like this:
```js
className={classNames(
  'inline-input',
  isValid(internalValue) || 'invalid',
  disabled && 'disabled',
  textAlignLeft && 'tw-pl-2 tw-text-left'
  className
)}
```

src/InlineInput/NumberInlineInput.tsx
```diff
@@ -77,21 +80,32 @@ const changeValueStepwise = (
...
+            className={`${className} number-inline-input`}
             disabled={disabled}
             value={String(value)}
-            isValid={newValue => isValid(Number(newValue), range)}
+            isValid={newValue => {
+                const validity = isValid(Number(newValue), range);
+                if (valueIsValidCallback != null) {
+                    // Then propagate the validity back to parent
+                    valueIsValidCallback(validity);
+                    return true;
```
kylebonnici, line 105: return valueIsValidCallback(validity);

kylebonnici: Also keep in mind  what we discussed i.e.

onValidationChanged?: (value: boolean) => void;
autoStyleInvalid: boolean;

kylebonnici (approved):

kylebonnici (approved):

PR #794, opened 2023-10-19
Feat/improve azure analytics

kylebonnici:


Changelog.md
```diff
@@ -7,6 +7,13 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 121 - UNRELEASED
+
+### Changed
+
+-   Analytic events names of prod and dev are now distinct with a prefix of
+    `Prod:` or`Dev:`
```
aadnekar, line 15: **Analytics** is the brand name used by Google for their Telemetry. I suggest that we try to refer to it as generically as possible, and therefore use Telemetry events, instead of Analytics events. The Microsoft equivalent to Google Analytics, I believe is called Azure Monitor Insights.

Suggestion:
```md
Telemetry events have distrinct prefixes `Prod:` and `Dev:`, where `Prod`: refers to all the sources except the `local` source in  nRF Connect for Desktop.`
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nordicsemiconductor/pc-nrfconnect-shared",
-    "version": "120.0.0",
+    "version": "121.0.0",
```
aadnekar, line 3: Can we sync this release with the PR I have already queued from v121, #793 ?

src/utils/usageData.ts
```diff
@@ -140,26 +144,25 @@ export const reset = () => {
...
+const sendEvent = ({ action, metadata }: EventAction) => {
     const isSendingUsageData = getIsSendingUsageData();
 
     if (isSendingUsageData && insights !== undefined) {
-        logger?.debug(`Sending usage data ${action} ${label}`);
-        insights.trackEvent({
-            name: action,
-            properties: label ? { label } : undefined,
-        });
+        logger?.debug(`Sending usage data ${action}`);
```
aadnekar, line 151: As it's a debug logger you could also include the full metadata.
```suggestion
        logger?.debug(`Sending usage data ${action}: ${JSON.stringify(metadata)}`);
```

aadnekar (approved):
Looks good, but please address the comments I've added 🐕‍🦺🚶

src/App/App.tsx
```diff
@@ -117,6 +102,14 @@ const ConnectedApp: FC<ConnectedAppProps> = ({
         action: openWindow.openLauncher,
     });
 
+    useEffect(() => {
+        paneName.current = allPanes.map(({ name }) => name);
+    }, [allPanes]);
+
+    useEffect(() => {
+        usageData.sendPageView(paneName.current[currentPane]);
+    }, [currentPane]);
```
boundlesscalm, line 111: I don't see the point in adding a ref.
Can't you just do 
```suggestion
    useEffect(() => {
        usageData.sendPageView(allPanes[currentPane].name));
    }, [currentPane]);
```

src/Device/DeviceSelector/DeviceSelector.tsx
```diff
@@ -62,6 +64,17 @@ export default ({
...
+            (): AppThunk => (_, getState) => {
+                const device = getState().device.selectedDevice;
+                if (device) {
+                    usageData.sendUsageData(
+                        'device deselected ',
+                        simplifyDeviceForLogging(device)
+                    );
+                }
+            }
+        );
```
boundlesscalm, line 77:
```suggestion
    const doDeselectDevice = useCallback((device: Device) => {
          usageData.sendUsageData(
              'device deselected ',
              simplifyDeviceForLogging(device)
          );
```

Alternatively, you can have it as an optional parameter to avoid the "currentDevice is of type Device | undefined" when calling it further down the file.

src/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -97,7 +91,10 @@ class ErrorBoundary extends React.Component<
             return children;
         }
 
-        const appDisplayName = appName || packageJson().displayName;
+        const appDisplayName =
+            appName ||
+            ((packageJson() as AppPackageJson).displayName ??
+                packageJson()?.name);
```
boundlesscalm, line 97: Isn't this just an x OR y OR z? Why are the last 2 options treated differently.

```suggestion
            appName ||
            (packageJson() as AppPackageJson).displayName ||
            packageJson()?.name;
```

package.json
```diff
@@ -107,12 +108,13 @@
         "redux": "4.2.1",
         "redux-devtools-extension": "2.13.9",
         "redux-thunk": "2.4.2",
+        "remove": "^0.1.5",
```
boundlesscalm, line 111: This seems to have been added by accident?
I can't find any usage from it in the code.

package.json
```diff
@@ -107,12 +108,13 @@
...
         "rimraf": "3.0.2",
         "roboto-fontface": "0.10.0",
         "semver": "7.3.8",
         "serialport": "^10.5.0",
         "shasum": "1.0.2",
-        "systeminformation": "5.17.12",
+        "systeminformation": "^5.17.12",
```
boundlesscalm, line 117: Any specific reason for this change?

src/App/App.tsx
kylebonnici: No as this will log access page views when we add the about and the feedback form to the `allPanes`

boundlesscalm: Not sure what you're trying to say.
This useEffect is triggered once on startup with both your or my approach.
No additional useEffect call is run because the `log` or `feedback` panes are added to `allPanes`.

package.json
kylebonnici: nop

datenreisender:
When in the renderer process we use an account/client ID to identify a certain user. We do not do that in the main process because AFAICS the library does not provide the possibility. Might this bring problems to the telemetric data if some have an ID and other not? Or should we remove the ID from the renderer process also?

datenreisender:
So far apps opted into sending telemetry with the `reportUsageData` property in the `App` component. With these changes all apps will send telemetry, including apps by third parties. Can we keep or restore the previous behaviour?

src/App/App.tsx
datenreisender: I also think the ref `paneName` is not needed.

PR #817, opened 2023-11-07
Feat: auto select device with --comPort

kylebonnici:


ipc/openWindow.ts
```diff
@@ -12,8 +12,26 @@ const channel = {
...
+): device is OpenAppOptionsDevicePort =>
+    (device as OpenAppOptionsDevicePort).serialPortPath !== undefined;
+
+export type OpenAppOptionsDeviceSN = { serialNumber: string };
+
+export type OpenAppOptionsDevicePort = { serialPortPath: string };
+
+export type OpenAppOptionsDevice =
+    | OpenAppOptionsDeviceSN
+    | OpenAppOptionsDevicePort;
```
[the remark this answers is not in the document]
aadnekar: The example below will ultimately do the same thing using Generic :) 

```typescript
interface OpenAppOptions {
    device?: 
        | { serialNumber: string; }
        | { serialPortPath: string; };
}

const getOpenAppOptions = <Device extends OpenAppOptions>(device: Device) => {
    return device;
}
```

aadnekar (approved):
LGTM, but please review my typescript type narrowing suggestion :)

ipc/openWindow.ts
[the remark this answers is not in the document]
aadnekar: I made some test cases that illustrates how the type checking will work, and I think it's similar to how you wrote them. I'm a bit puzzled if we want to avoid the cases where you provide both `serialNumber` and `serialPortPath` or if we want to accept them?

```typescript
// ✅
const value1 = getOpenAppOptions({ device: { serialNumber: '0909090909' } });
type expected1 = { device: { serialNumber: string}};
type test1 = Expect<Equal<typeof value1, expected1>>

// ✅
const value2 = getOpenAppOptions({ device: { serialPortPath: '/dev/ttyACM0' } });
type expected2 = { device: { serialPortPath: string}};
type test2 = Expect<Equal<typeof value2, expected2>>

// ✅
const value3 = getOpenAppOptions({ device: undefined });
type expected3 = { device: undefined};
type test3 = Expect<Equal<typeof value3, expected3>>

// ✅
const value4 = getOpenAppOptions({});
type expected4 = {};
type test4 = Expect<Equal<typeof value4, expected4>>


// Does not work, because it expects device.
// @ts-expect-error ==> ❌
const invalid1 = getOpenAppOptions();

// This is actually valid: ✅
const invalid2 = getOpenAppOptions( { device: { serialPortPath: '0909090909', serialNumber: '/dev/ttyACM0' } } );
```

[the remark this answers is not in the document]
kylebonnici: Use case I have can be found here 
https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/blob/fix/auto-connect-single-instance/src/main/windows.ts#L126

main/index.ts
```diff
@@ -46,4 +46,13 @@ export {
 } from '../ipc/schema/packageJson';
 
 export { type OverwriteOptions } from '../ipc/serialPort';
-export { type OpenAppOptions } from '../ipc/openWindow';
+export type {
+    OpenAppOptions,
+    OpenAppOptionsDevice,
+    OpenAppOptionsDeviceSN,
+    OpenAppOptionsDevicePort,
```
datenreisender, line 53: Only `OpenAppOptions` needs to be exported, the three later ones are not really used by the launcher.

ipc/openWindow.ts
```diff
@@ -12,8 +12,26 @@ const channel = {
...
+export const isOpenAppOptionsDevicePort = (
+    device: OpenAppOptionsDevice
+): device is OpenAppOptionsDevicePort =>
+    (device as OpenAppOptionsDevicePort).serialPortPath !== undefined;
+
+export type OpenAppOptionsDeviceSN = { serialNumber: string };
+
+export type OpenAppOptionsDevicePort = { serialPortPath: string };
+
+export type OpenAppOptionsDevice =
```
datenreisender, line 29: Just like in `main/index.ts` the three types `OpenAppOptionsDevice*` do not really have to be exported but could be kept internal here.

src/Device/deviceLister.ts
```diff
@@ -367,10 +366,42 @@ export const startWatchingDevices =
...
+        )
+    );
+
+    if (serialNumber && serialPort) {
+        if (serialNumber.index > serialPort.index) {
+            return serialPort.device;
+        }
+
+        return serialNumber.device;
+    }
```
datenreisender, line 396: In the launcher you, as discussed, do not handle the case if both serial number and port are given (instead we print an error). Here you do handle the case, with the rule that the later command line argument takes precedence. 

Both would be fine for me: Either keep it here like it is or just ignore the case here (by removing those 7 lines above, which would effectively give the serial number precedence if the impossible case happens that both are given).

datenreisender (approved):
LGTM 👍

Left some comments, but those are no show-stoppers. We or you could also change those later, if we want to.

We now have the strings for the two command line switches `--comPort` and `--deviceSerial` thrice in the source code: One time here and in two files in the launcher. Would probably be nice to have it specified only ones, but this is also something we can consider later.

src/Device/deviceLister.ts
kylebonnici: So we have three cases here 

- user opens app from cli with -- --deviceSerial ... --comPort ... -> in this case we close app immidiatly

- user opens app from cli with -- --deviceSerial ... e.g. cellular monitor --> app will open and have the -- --deviceSerial  in its argv, if user then runs click the button to open serial terminal and that will need to pass device -- --comPort launcher is already open (with no window) and we need to make sure that we pass the `mergeAdditionalArguments` first to always use what ever the openApp needs to do dance the order above. 

- user opens app from cli with -- --deviceSerial ... --> app will open and have the -- --deviceSerial  in its argv, if user then runs -- --comPort launcher is already open (with no window) and we need to make sure that we pass the arguments from the second CLI instance first as that is what we need to use .... having said this I am unable to verify this last case ...

PR #824, opened 2023-11-10
Feat/extend nrfutil sandbox

kylebonnici:


nrfutil/sandbox.ts
```diff
@@ -276,47 +324,201 @@ export class NrfutilSandbox {
...
+    };
+
+    public execCommand = (
+        command: string,
+        args: string[],
+        onData: (data: Buffer, pid?: number) => void,
         onStdError: (data: Buffer, pid?: number) => void,
-        controller?: AbortController
+        controller?: AbortController,
+        filterEnv: (env: NodeJS.ProcessEnv) => NodeJS.ProcessEnv = env => env
```
boundlesscalm, line 510: How about renaming this and everywhere else to something like `editEnv`.
Since it allows you to do (and probably will be used for) more than just filtering the env.

nrfutil/sandboxTypes.ts
```diff
@@ -17,13 +17,11 @@ export type FeatureClassification =
     | 'nrf-external-confidential'
     | 'nrf-external';
 
-export type Task = {
+export type Task<T = void> = {
```
aadnekar, line 20: Why have you chosen `void` as the default type of `T`?

Typescript Documentation clearly state that void is only used to describe that a function returns nothing: https://www.typescriptlang.org/docs/handbook/2/functions.html#void.

I would think that this should either be `undefined` or `null`, and probably `null`, as `undefined` may give the assumption that it could at any point become defined? 🤔

nrfutil/sandboxTypes.ts
```diff
@@ -36,11 +34,10 @@ export type TaskBegin = {
 };
 
 export type TaskEnd<T = void> = {
```
aadnekar, line 36: Same as above, should probably not be `void`, but rather `null`.

package.json
```diff
@@ -115,6 +115,7 @@
         "shasum": "1.0.2",
         "systeminformation": "5.17.12",
         "tailwindcss": "3.3.2",
+        "tree-kill": "^1.2.2",
```
aadnekar, line 118: `tree-kill` has not been updates in 5 years, nov 2018, and should probably not be used.
There is a different project called `tree-kill-promise`, which is more recently updated: https://github.com/dword-design/tree-kill-promise.

Tree-kill-promise:
https://github.com/dword-design/tree-kill-promise
Distributed package seems well done 🤘
https://unpkg.com/browse/tree-kill-promise@3.0.14/

aadnekar (commented):
Looks okay to me, but I don't fully understand the changes 🫣
Please, check out my comments, as I think they should be addressed sooner rather than later 🕵️

PR #826, opened 2023-11-14
fix: process device in order as there are enumerated.

kylebonnici:


Changelog.md
```diff
@@ -7,6 +7,18 @@ This project does _not_ adhere to
...
 every new version is a new major version.
 
+## 130 - 2023-11-14
+
+### Fixed
+
+-   Ensure that all enumerated devices are always processed in order to avoid
+    calling:
+    -   nrfutil device-info multiple times for the same device
+    -   only call onSuccess once for waiting for device when rebooting
```
boundlesscalm, line 17:
```suggestion
-   Ensure that all enumerated devices are always processed in order to avoid:
    -   calling nrfutil device-info multiple times for the same device
    -   only calling onSuccess once when waiting for device when rebooting
```

Changelog.md
```diff
@@ -7,6 +7,18 @@ This project does _not_ adhere to
...
+-   `nrfutil device` call to `device-info` is called before notifying the app
+    with an onSelectedDevice. This is to ensure the operations to the device are
+    completed before the app possibly does any more action on it
```
boundlesscalm, line 20:
```suggestion
-   `nrfutil device` now calls `device-info` before notifying the app
    with an onSelectedDevice. This is to ensure the operations to the device are
    completed before the app possibly does any more action on it
```

PR #831, opened 2023-11-16
Make the Dropdown component generic

aadnekar:
Make it so that the DropDownItem type is inferred from the properties passed to the Dropdown component.

Some mocked use case to demonstrate:

```tsx
interface CustomValue {
    foo: string;
    bar: number;
}

const MyComponent = () => {
    const items: readonly DropdownItem<CustomValue>[] = someListOfSorts.map( ... );
    
    return <Dropdown
        selectedItem={items[0]}
        items={items}
        onSelect={ (item) => {
            console.log(item.foo) // ✅
            console.log(item.bar) // ✅
        }} />
};

```

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -16,18 +16,18 @@ export interface DropdownItem<T = string> {
     value: T;
 }
 
-export interface DropdownProps {
+export type DropdownProps<T> = {
     id?: string;
     label?: React.ReactNode;
-    items: DropdownItem[];
-    onSelect: (item: DropdownItem) => void;
+    items: DropdownItem<T>[];
+    onSelect: (item: DropdownItem<T>) => void;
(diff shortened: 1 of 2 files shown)


datenreisender (approved):

src/Dropdown/Dropdown.tsx
```diff
@@ -16,30 +16,32 @@ export interface DropdownItem<T = string> {
...
     numItemsBeforeScroll?: number;
     className?: string;
-}
+};
 
-export default ({
+export default <T,>({
     id,
     label,
+    defaultButtonLabel,
```
datenreisender, line 34:
```suggestion
    defaultButtonLabel = '',
```

By using a default value here, `defaultButtonLabel ?? ''` shouldn't be needed anymore.

datenreisender (approved):

PR #832, opened 2023-11-17
Fix metadata removal

datenreisender:
Looking at the documenation I found it hard to deduce, what `envelope.data.baseData` is supposed to contain. But looking at what is actually in there when `usageData.sendUsageData` is invoked from the main process in preparing an nrfutil sandbox I think that `removeAllMetadata` will be in `envelope.data.baseData.properties` and that the `name` will be in `envelope.data.baseData.name`.


kylebonnici (approved):
Documentation is very lacking. Thanks for sotting this :)

PR #860, opened 2024-01-10
Keep persisted `isLoggingVerbose` state up-to-date during runtime

boundlesscalm:


src/About/SupportCard.tsx
```diff
@@ -73,6 +76,7 @@ export default () => {
                             label="VERBOSE LOGGING"
                             onToggle={isToggled => {
                                 NrfutilDeviceLib.setVerboseLogging(isToggled);
+                                persistIsLoggingVerbose(isToggled);
```
datenreisender, line 79: S: Instead of persisting the setting here it would be more aligned with the rest of our codebase if this is done in the reducer (e.g. in the `appLayout` slice we also do it that way)

src/About/SupportCard.tsx
```diff
@@ -83,7 +87,7 @@ export default () => {
                         <Button
                             variant="secondary"
                             onClick={() => {
-                                persistIsLoggingVerbose(true);
+                                removeIsLoggingVerboseResetHandler();
```
datenreisender, line 90: W: Instead of expressing the “how” with `removeIsLoggingVerboseResetHandler` I would name this to more express the “what/why”, e.g. with `doNotResetLoggingOnRestart`.

src/utils/persistentStore.ts
```diff
@@ -181,5 +176,26 @@ export const persistLogVisible = (visible: boolean) =>
...
+    getAppSpecificStore<SharedAppSpecificStoreSchema>().get(
+        'isLoggingVerbose',
+        false
+    );
+
+const resetIsLoggingVerbose = () =>
+    getAppSpecificStore<SharedAppSpecificStoreSchema>().set(
+        'isLoggingVerbose',
+        false
+    );
```
datenreisender, line 195: S: Reuse `persistIsLoggingVerbose` from above:
 ```suggestion
const resetIsLoggingVerbose = () => persistIsLoggingVerbose(false);
```

datenreisender (approved):
Looks good, some suggestions, following [MoSCoW](https://dev.to/allthecode/moscow-the-best-code-review-technique-you-re-not-using-2b0e).

src/About/SupportCard.tsx
boundlesscalm: I considered this, but felt it was a side effect I didn't want in the slice.
However it makes more sense there, given that it can be called from other places.

Should the line above `NrfutilDeviceLib.setVerboseLogging()` also be moved to the slice as it follows the same idea?
Both the slice and the verbose logging toggle will always exist in the `App` context and should always be set together.

datenreisender: Persisting the value in the reducer is a side-effect. But it is the way we do it everywhere at the moment, so I prefer to stick with the pattern. At least for now. 

A way to solve this while adhering to the Redux principles might be to introduce a middleware for persisting these values, BTW. Another solution would be to always use effects for setting persisted values. But in both cases we should also implement reading the initial values from the persisted store different from how we do it now. But the current pattern doesn't cause us too many problems, so I am not too keen to do that. 

I wouldn't move `NrfutilDeviceLib.setVerboseLogging()` into reducer because it is also a side-effect. And even though we have the persistence in there, I also do not want to move even more side-effects in there.

boundlesscalm: Completely understandable.

I am just worried that we may at some point use the reducer other places to enable/disable verbose logging and not remember that we also have to explicitly set verbose logging for nrfutil.
It is something that works hand in hand with this setting after all.
I will still change it as you described, but what about adding a comment in the reducer to make sure this doesn't turn into primal knowledge?

datenreisender: A comment is fine. Another option is to create an effect which combines invoking `NrfutilDeviceLib.setVerboseLogging()` and dispatching `setIsLoggingVerbose()`.

boundlesscalm: I added an effect.
For now I kept it in the same file to avoid confusion between 2 different `setIsLoggingVerbose` actions.

datenreisender (approved):

PR #861, opened 2024-01-10
feat: export all package.json listed nrfutil modules from shared

boundlesscalm:
This PR does a few things, all of which are required to allow shared to log the nrfutil module version which are in use.

1. All modules listed in the package.json of an app are accessible from shared and are lazy-loaded upon the first `get`
2. To avoid cyclic dependencies and for seperation of concern, the types of `nrfutil` and `nrfutil device` were split into separate `index.ts` files
3. Since all nrfutil modules are available in shared, enabling verbose logging now works for all modules (compared to only nrfutil device like previously). Depends on https://github.com/NordicSemiconductor/pc-nrfconnect-shared/pull/860.

### DIFF
--- a/nrfutil/device/logLibVersions.ts
+++ b/nrfutil/device/logLibVersions.ts
@@ -7,19 +7,20 @@
 import { spawn } from 'child_process';
 import os from 'os';
 
-import logger from '../../src/logging';
 import {
     describeVersion,
     getExpectedVersion,
     resolveModuleVersion,
 } from '../moduleVersion';
 import { ModuleVersion, SubDependency } from '../sandboxTypes';
+import { getNrfutilLogger } from '../nrfutilLogger';
 
 const log = (description: string, moduleVersion?: SubDependency | string) => {
+    const logger = getNrfutilLogger();
     if (moduleVersion == null) {
-        logger.warn(`Unable to detect version of ${description}.`);
+        logger?.warn(`Unable to detect version of ${description}.`);
     } else {
-        logger.info(
+        logger?.info(
             `Using ${description} version: ${describeVersion(moduleVersion)}`
         );
     }
@@ -68,6 +69,7 @@ const checkJLinkArchitectureOnDarwin = async () => {
 };
 
 export default async (moduleVersion: ModuleVersion) => {
+    const logger = getNrfutilLogger();
     try {
         const dependencies = moduleVersion.dependencies;
 
@@ -81,12 +83,12 @@ export default async (moduleVersion: ModuleVersion) => {
         if (jlinkVersion) {
             const result = getExpectedVersion(jlinkVersion);
             if (!result.isExpectedVersion) {
-                logger.warn(
+                logger?.warn(
                     `Installed JLink version does not match the expected version (${result.expectedVersion})`
                 );
             }
         } else {
-            logger.warn(
+            logger?.warn(
                 `JLink is not installed. Please install JLink from: https://www.segger.com/downloads/jlink`
             );
         }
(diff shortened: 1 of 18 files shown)


datenreisender:
Since this depends on https://github.com/NordicSemiconductor/pc-nrfconnect-shared/pull/860 (as you wrote) I would usually not base this PR on `main` but instead on the branch `feat/persisted-verbose-logging` of that PR. 

Or do you prefer to keep PRs against `main` since this is where the code will end up?

For me basing this PR on `feat/persisted-verbose-logging` would make reviewing it independent of that other PR a bit easier, but I can achieve the same by looking at the diff manually at https://github.com/NordicSemiconductor/pc-nrfconnect-shared/compare/feat/persisted-verbose-logging...feat/shared-controlled-nrfutil-modules

At the end it doesn't change a lot, as soon as #860 is merged, the situation will be the same: 
- If this PR is against `main`, then the two commits from #860 should just disappear from this PR. 
- If this PR is against `feat/persisted-verbose-logging`, then this PR would automatically change to be against `main` as soon as #860 is merged.

boundlesscalm:
I wasn't aware that the PR would automatically be redirected against main, that makes this so much nicer.
I'll re-target this against the verbose logging branch. Thanks!

nrfutil/modules.ts
```diff
@@ -0,0 +1,166 @@
...
+            acc.then(async () => {
+                if (val.isInitialised()) {
+                    moduleSandboxes.push(await val.get());
+                }
+            }),
+        Promise.resolve()
+    );
+
+    return moduleSandboxes;
+};
```
datenreisender, line 142: From what I understand the `reduce` serialises the promises. But what is the advantage of this? Creating them in parallel seems simpler and potentially faster:
```ts
const getAllIninitialisedModules = () =>
    Promise.all(
        Object.values(modules)
            .filter(module => module.isInitialised)
            .map(module => module.get())
    );
```

nrfutil/modules.ts
```diff
@@ -0,0 +1,166 @@
...
+    const moduleVersions: ModuleVersion[] = [];
+    (await getAllIninitialisedModules()).reduce(
+        (acc, moduleSandbox) =>
+            acc.then(async () => {
+                moduleVersions.push(await moduleSandbox.getModuleVersion());
+            }),
+        Promise.resolve()
+    );
+    return moduleVersions;
+};
```
datenreisender, line 166: Like above, parallelising this seems easier:
```ts
export const getAllModuleVersions = async () =>
    Promise.all(
        (await getAllIninitialisedModules()).map(module =>
            module.getModuleVersion()
        )
    );
```

nrfutil/device/logLibVersions.ts
```diff
@@ -101,15 +103,15 @@ export default async (moduleVersion: ModuleVersion) => {
...
                 );
-                logger.warn(
+                logger?.warn(
                     `Please install JLink: https://www.segger.com/downloads/jlink/JLink_MacOSX_V788j_universal.pkg`
                 );
             }
         }
     } catch (error) {
-        logger.logError('Failed to get the library versions', error);
+        logger?.error('Failed to get the library versions', error);
```
datenreisender, line 115: Please check whether this is correct, I assume the change from `logError` to `error` is a mistake: [logError](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/424faf72d87bd96ebb95fce0aba2132587d30f7e/src/logging/index.ts#L96-L101) is a function which outputs the second argument. With `error` the second argument is considered [“meta”](https://github.com/winstonjs/winston/tree/master#streams-objectmode-and-info-objects) and probably discarded.

datenreisender (approved):
Looks good.

I pushed three suggestions to the branch [suggestions/feat/shared-controlled-nrfutil-modules](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/compare/feat/shared-controlled-nrfutil-modules...suggestions/feat/shared-controlled-nrfutil-modules). Feel free to take or dismiss them.

Left three more comments where I was unsure whether I was right but it would be good if you take a look at it.

nrfutil/device/logLibVersions.ts
boundlesscalm: You're correct.
I had not found `logError` (though I obviously didn't check very thoroughly) and this gave me type errors. I'll see whether I will just string format this or fix the types.

boundlesscalm: Decided to reimplement the logError explicitly here.
The original logError was part of the `SharedLogger` which I did not want to reference here as we should only rely on the `nrfutilLogger`.

boundlesscalm:
@datenreisender Thanks for the suggestions!
Implemented all of them and merged your branch.

PR #872, opened 2024-01-26
ci: don't fail label flow in documentation pr

boundlesscalm:


.github/workflows/labels.yml
```diff
@@ -13,7 +13,8 @@ jobs:
           - name: 'Check for required labels'
             if: >
                 contains(github.event.pull_request.labels.*.name, 'doc required') == false &&
-                contains(github.event.pull_request.labels.*.name, 'doc not required') == false
+                contains(github.event.pull_request.labels.*.name, 'doc not required') == false &&
+                contains(github.event.pull_request.labels.*.name, 'documentation') == false
```
datenreisender, line 17: Instead of comparing with `false` could we also treat this as proper booleans?
```yml
if: >
     ! contains(github.event.pull_request.labels.*.name, 'doc required') &&
     ! contains(github.event.pull_request.labels.*.name, 'doc not required') &&
     ! contains(github.event.pull_request.labels.*.name, 'documentation')
```

Not sure about whether this needs to be wrapped in `${{}}`. On https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution it says 

> You must always use the `${{ }}` expression syntax or escape with `''`, `""`, or `()` when the expression starts with `!`, since `!` is reserved notation in YAML format.

But I think here we start the expression with `>`, so I hope it is not needed.

datenreisender (approved):

.github/workflows/labels.yml
boundlesscalm: Not sure.
I'll just create a pr and reference this branch to test the label action.

PR #874, opened 2024-01-26
Linter should ignore resources

aadnekar:
Would solve issues where resource files should not be auto formatted:
e.g. https://github.com/NordicSemiconductor/pc-nrfconnect-cellularmonitor/pull/419

### DIFF
--- a/config/eslintrc.js
+++ b/config/eslintrc.js
@@ -24,6 +24,7 @@ module.exports = {
     ],
     ignorePatterns: [
         'doc/docs/',
+        'resources/',
         'package-lock.json',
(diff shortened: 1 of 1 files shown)


datenreisender:
I suggest that we instead do it [like in the launcher](https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/blob/2efb8cb517156b4ec7093199bd71d7e7623ecc54/.eslintrc.js#L11) and specify just for the CM to exclude certain JSON files (not sure whether we want to exclude only the trace DB definitions or also the `config_v3.json`).

PR #884, opened 2024-02-06
NCD-613: BCTL Add getBoardControllerConfig nrfutil commands

cybic:
Add commands to read config from board controller firmware


kylebonnici:
Looks good :) just one thing, open ticket with the maintainer of the nrfutil developers for this command to eventually add cli params to allow this to become like the other commands and this also eliminate need for us to maintain the json type

kylebonnici (approved):

PR #899, opened 2024-03-04
Feat: implement group with no bootstrap dependency

kylebonnici:


src/Groups/Group.tsx
boundlesscalm, line 1: why is the folder called `Groups`

boundlesscalm (approved):

src/Groups/Group.tsx
kylebonnici: becase I added and `s` ... :P my thinking was to be able to add other suff in the future like gripped grapes where if one is opened the others close etc... bit I can rename

PR #914, opened 2024-04-09
Feat: Add publishing date for non-official sources

KievDevel:
Motivation: We use sha sum to calculate if build is updated, but sometimes I'm missing timestamps when same version is released over and over again, ending up testing wrong version of the build.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,12 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 172.0.0 - UNRELEASED
+
+### Added
+
+-   Publish timestamp for nordic-publish for non-official sources
+
 ## 171.0.0 - 2024-04-04
 
 ### Added
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,12 +7,6 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
-## 175.0.0 - UNRELEASED
-
-### Added
-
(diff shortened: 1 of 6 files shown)


ipc/apps.ts
```diff
@@ -34,6 +34,7 @@ interface Installed {
     repositoryUrl?: UrlString;
     html?: string;
     installed: {
+        publishTimestamp?: string;
```
datenreisender, line 37: Whether you also need to add this, depends on what you want to do with the timestamp (which I do not know).

scripts/nordic-publish.ts
```diff
@@ -269,6 +270,7 @@ const updateLegacyAppInfo = (appInfo: LegacyAppInfo, app: App) => {
             ...appInfo.versions,
             [app.version]: {
                 dist: {
+                    publishTimestamp: new Date().toISOString(),
```
datenreisender, line 273: This isn't really needed but also causes no harm.

ipc/apps.ts
KievDevel: Will revisit this. Initially I want to display timestamp of the already installed version of the app. Seems like the right place to add.

datenreisender: 👍

datenreisender (approved):

Changelog.md
```diff
@@ -7,6 +7,12 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 175.0.0 - UNRELEASED
+
+### Added
+
+-   Publish timestamp for nordic-publish for non-official sources.
+
```
boundlesscalm, line 15: Move into v174

datenreisender (approved):

PR #936, opened 2024-08-23
Handle "Failed to get the library versions" if error is undefined

KievDevel:
**What it does?**
Removes "undefined" from the error message

image-20240823-002216

---
Code I don't like much, but it definitely doesn't break existing one :)

### DIFF
--- a/nrfutil/device/logLibVersions.ts
+++ b/nrfutil/device/logLibVersions.ts
@@ -113,11 +113,10 @@ export default async (moduleVersion: ModuleVersion) => {
             }
         }
     } catch (error) {
-        let message = '';
-       logger?.error(
-             `Failed to get the library versions${
-                 error != null ? `: ${describeError(error)}` : ''
-             }`
-         );
+        logger?.error(
+            `Failed to get the library versions${
+                error != null ? `: ${describeError(error)}` : ''
+            }`
+        );
     }
 };
--- a/nrfutil/device/logLibVersions.ts
+++ b/nrfutil/device/logLibVersions.ts
@@ -108,7 +108,7 @@ export default async (moduleVersion: ModuleVersion) => {
                     `It looks like you have installed JLink using ${JLinkInstallerVersion}, but currently we only support their Universal Installer for your system.`
                 );
                 logger?.warn(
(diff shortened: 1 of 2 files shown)


nrfutil/device/logLibVersions.ts
```diff
@@ -108,15 +108,19 @@ export default async (moduleVersion: ModuleVersion) => {
...
-                error == null ? `: ${describeError(error)}` : ''
-            }`
-        );
+        let message = '';
+        if (error == null) {
+            message = describeError(error);
+        }
+        if (error === undefined) {
+            message = '';
+        }
```
kylebonnici, line 122: This is not the same as 

`error == null ? `: ${describeError(error)}` : ''`

error == null is true for null and undefined as we are using == and not ===

regardless I don't understand why we use describeError(error) when error is null or undefined IMO it should be 

`logger?.error(
             `Failed to get the library versions${
                 error != null ? `: ${describeError(error)}` : ''
             }`
         );`

kylebonnici, line 122:
```suggestion
       logger?.error(
             `Failed to get the library versions${
                 error == null ? `: ${describeError(error)}` : ''
             }`
         );
```

nrfutil/device/logLibVersions.ts
```diff
@@ -108,15 +108,19 @@ export default async (moduleVersion: ModuleVersion) => {
...
+
+        logger?.error(`Failed to get the library versions${message}`);
```
kylebonnici, line 124:
```suggestion
       logger?.error(
             `Failed to get the library versions${
                 error != null ? `: ${describeError(error)}` : ''
             }`
         );
```

kylebonnici (approved):
On minor commet otherwise merge :) thanks for the PR :) :) :)

kylebonnici:
One minor cmt and merge.... do not forget to update the changelog and the version

nrfutil/device/logLibVersions.ts
KievDevel: didn't do that because it changes existing logic.
"regardless I don't understand why we use describeError(error) when error is null or undefined IMO it should be as the cmt below" - yes I too did not understand that but decided not to change.

null & undefined are handled in describeError() and converted to strings there, but that doesn't make sense in this use case

PR #943, opened 2024-10-02
feat: add offline devices

boundlesscalm:
Still requires hover text change for top left button
Screenshot from 2024-11-20 13-37-30


kylebonnici (approved):
Looking very nice :)

PR #949, opened 2024-10-17
Feat/add dropdown data constructors

tsvetelinpetrov:


kylebonnici (approved):
Looks good

- Just add the change log and increment the version 
-consider renaming the file to make it clearer that this is to be used as helper for dropdown.
- Add the new functions to the index.ts so apps can import

PR #961, opened 2025-01-08
Improve shortcuts modal layout & look

KievDevel:
BEFORE:

image

AFTER:

image


ketile:
The change to default dialog looks good. 

In my opinion we should keep the horizontal dividers and the extra vertical spacing from the original. The original left/right alignment of title and icon should also be kept. These changes will make the list look less cluttered.

datenreisender (dismissed):
Agree with Ketil on the stylistic aspects. The technical aspects are fine.

KievDevel:
@ketile reason for the first implementation is that I don't like distance between label and button shortcuts, it feels too far away with decreased ability to quickly read :( 
Here is an updated version based on the comments:

image

ketile:
I recommend removing the colons. Might also use a more narrow version of the dialog if one exists.

KievDevel:
This was my first attempt, but existing modal is too narrow. Something in between would be perfect, but I didn't want to create another option of the modal for the single use case

image

PR #962, opened 2025-01-08
Feedback modal: generic handle error

KievDevel:
Generic error message (API does not respond / network error): 

image

### DIFF
--- a/src/About/Feedback.tsx
+++ b/src/About/Feedback.tsx
@@ -79,7 +79,7 @@ const FeedbackDialog = ({
                             }}
                             disabled={feedback === '' || sendingFeedback}
                         >
-                            Send feedback
+                            Send
                         </DialogButton>
                     )}
                     <DialogButton onClick={onClose}>Close</DialogButton>
@@ -100,7 +100,8 @@ const FeedbackDialog = ({
                         <p>
                             We value your feedback and any ideas you may have
                             for improving nRF Connect for Desktop applications.
-                            Use the form below.
+                            Please write below how we can make this application
+                            better.
                         </p>
                         <p>
                             We only collect the following information when you
--- a/src/About/Feedback.tsx
+++ b/src/About/Feedback.tsx
@@ -45,6 +45,7 @@ const FeedbackDialog = ({
 
     const onClose = () => {
         setFeedback('');
+        setErrorMessage('');
         setSendingFeedback(false);
         setSayThankYou(false);
         setSelectedCategory(categoryItems ? categoryItems[0] : undefined);
--- a/src/About/Feedback.tsx
+++ b/src/About/Feedback.tsx
@@ -27,6 +27,7 @@ const FeedbackDialog = ({
     const [feedback, setFeedback] = useState('');
     const [sayThankYou, setSayThankYou] = useState(false);
     const [sendingFeedback, setSendingFeedback] = useState(false);
+    const [errorMessage, setErrorMessage] = useState('');
 
     const categoryItems = useMemo(() => {
         if (!categories?.length) return undefined;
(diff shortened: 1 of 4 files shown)


greg-fer:
Maybe:

> "Use the form below" > "Please write below how we can make this application better."
 
To narrow down the scope of the feedback we want (this app).

And:
 
> "Send feedback" > "Send"

src/About/Feedback.tsx
```diff
@@ -124,6 +132,11 @@ const FeedbackDialog = ({
                                     value={feedback}
                                     onChange={e => setFeedback(e.target.value)}
                                 />
+                                {errorMessage ? (
+                                    <p className="tw-mb-0 tw-mt-1 tw-text-right tw-text-xs tw-text-red">
+                                        {errorMessage}
+                                    </p>
+                                ) : null}
```
boundlesscalm, line 139: Maybe just me but I feel like a ternary expression makes this a bit harder to read.
```suggestion
                                {!!errorMessage && (
                                    <p className="tw-mb-0 tw-mt-1 tw-text-right tw-text-xs tw-text-red">
                                        {errorMessage}
                                    </p>
                                )}
```

boundlesscalm (approved):
LGTM

greg-fer (approved):

src/About/Feedback.tsx
KievDevel: I do this way because some engines can produce bugs because of expression rendering false / undefined.
Remix JS and React Native have this issue so I better be safe :) Our setup seems ok, but that's more as a new habit.

With the readability part I agree

ketile:
- In addition to saying "an error occurred" we should state what the error is as best we can.
- If "try again" does not solve the issue we should provide a link to support on DevZone.
- Consider adding some text that explicitly states feedback will not be answered.

@boundlesscalm has made a error/warning component for the Quickstart application which I think would be a good fit for this use case.

src/About/Feedback.tsx
datenreisender: Instead of Mias double negation, a simple negation and `||` should also work:

```suggestion
                                {!errorMessage || (
                                    <p className="tw-mb-0 tw-mt-1 tw-text-right tw-text-xs tw-text-red">
                                        {errorMessage}
                                    </p>
                                )}
```
(In my head I am reading this as “Either there is no error message or you output the error message”)

But I am fine with all three variants.

src/About/Feedback.tsx
```diff
@@ -136,13 +149,18 @@ const FeedbackDialog = ({
 const handleFormData = async (
     feedback: string,
     setResponse: (response: boolean) => void,
-    category?: string
+    category?: string,
+    onError?: () => void
```
datenreisender, line 153: - What is the purpose of this callback, is there already a plan to use it somewhere (for what?)? 
- Would it be helpful to also pass the error to it?
- Should its addition be mentioned in the changelog?

datenreisender (approved):

KievDevel:
@ketile did you mean this one?

- Added a par. that feedback will not be answered.
- Slightly updated an error message. Error most likely on user side like being offline. rare edge case would be that feedback server API does not work. In both cases we're unable to submit.
- Used component from the quickstart 

image

src/About/Feedback.tsx
KievDevel: 1) currently there is no way to know that any error happened outside of the handleFormData(). I could throw error next and catch later where the hook is called, but this sounds unsafe if hook is used elsewhere.
2) True, but looks unnecessary for this use case, can be added later if needed?
3) It is optional, doesn't break anything and most likely not needed for others, so I think no? Here I'm not sure

greg-fer (changes requested):
Please remove "Feedback will not be answered.".

- The feedback is by definition one way: "the transmission of evaluative or corrective information about an action, event, or process to the original or controlling source"
- The current phrasing of the sentence is direct and harsh.
- Including this message - which I haven't seen in any other feedback form elsewhere - can deter users from using this form.

ketile:
@KievDevel Looks great with latest changes

PR #967, opened 2025-01-28
Add mock for appinsights

KievDevel:
Reason: hide warning ` ApplicationInsights:Cannot load @azure/core-auth package. This package is required for AAD token authentication. It's likely that your node.js version is not supported by the JS Azure SDK.` from the library caused by jest.requireActual()

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,10 @@ every new version is a new major version.
 
 -   `selectedVirtualDevice` export.
 
+### Changed
+
+-   `applicationinsights` is now mocked for test runs.
+
 ### Fixed
 
 -   Show log shortcut now works every time it is called.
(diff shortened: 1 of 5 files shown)


datenreisender (approved):
Thanks, this annoyed me too, I just didn't get around to fixing it.

There isn't really a need for `mocks/applicationinsightsMock.ts` since we already have `mocks/emptyMock.ts`. I added a commit which uses the latter instead. If you prefer to keep `mocks/applicationinsightsMock.ts`, e.g. because you prefer the expressiveness it adds, then feel free to remove my commit again.

We could probably also remove the `pc-nrfjprog-js|nrf-device-setup|usb|` from `config/jest.config.js`.

KievDevel:
@datenreisender  Captured in https://nordicsemi.atlassian.net/browse/NCD-1174

PR #968, opened 2025-01-28
Add highlight color

KievDevel:
as use case is quite specific, don't think should be added in other places

matching PR: https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/pull/1058

### DIFF
--- a/styles.scss
+++ b/styles.scss
@@ -17,6 +17,8 @@ $primary-darkened: color.scale($nordic-blue, $lightness: -10%);
 $secondary: $white;
 $accent: $blue-slate;
 
+$highlight: #F0FBFC;
+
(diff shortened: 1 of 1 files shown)


KievDevel:
Change, so minor feels like not worth mentioning in the changelog

datenreisender (commented):
Thanks for linking the PR in the launcher, making it easy for me to understand the context.

I added https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/pull/1058#issuecomment-2624155196 there. Currently I assume that we should not add the color here in shared, but maybe that other PR is the better place to discuss this.

KievDevel:
Closed based on comment in the launcher repo

PR #971, opened 2025-02-06
Add workflows for the migration to Artifactory

datenreisender:
Part of [NCD-795](https://nordicsemi.atlassian.net/browse/NCD-795):

Add workflows for the migration to Artifactory. 

The goal is to build the apps in a GitHub Action (no more Azure pipeline for that), be automatically released to the source "latest" in Artifactory for pushes to `main` and enable releasing to any source (including "official" on Artifactory).

For this, the files `build.yml`, `release.yml`, and `release-latest.yml` need to be copied from https://github.com/NordicSemiconductor/pc-nrfconnect-boilerplate/tree/main/.github/workflows to `.github/workflows` of each app. (The folder is not yet really there in `pc-nrfconnect-boilerplate` because of a chicken and egg problem with this PR. But you can already see it in https://github.com/NordicSemiconductor/pc-nrfconnect-boilerplate/blob/add_artifactory_github_actions/.github/workflows/.)

I tested the workflows on the RSSI project: https://github.com/NordicSemiconductor/pc-nrfconnect-rssi/actions

This also adds a workflow which tests shared in GitHub Actions, so that we also move towards deprecation of Azure pipelines for shared.


.github/workflows/build-app.yml
```diff
@@ -0,0 +1,35 @@
+name: Build App
+
+on:
+    workflow_call:
+        inputs:
+            ref:
+                type: string
```
boundlesscalm, line 7: Is there a reason to not make this required?

.github/workflows/release-app.yml
```diff
@@ -0,0 +1,39 @@
+name: Release
+
+on:
+    workflow_call:
+        inputs:
+            ref:
+                type: string
```
boundlesscalm, line 7: Same as the other comment

.github/workflows/test-shared.yml
```diff
@@ -0,0 +1,21 @@
...
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4.0.0
+              with:
+                  node-version: 22
+                  cache: 'npm'
+            - run: npm ci
+            - run: npm run check
+            - run: npm test
+            - run: npm pack
```
boundlesscalm, line 21: Is packing necessary here?

.github/workflows/build-app.yml
```diff
@@ -0,0 +1,35 @@
...
+            force_pack:
+                type: boolean
```
boundlesscalm, line 9: Wondering where the comes from (at least I couldn't see it being passed in the workflow using this in RSSI)

boundlesscalm (approved):
Left some comments but looks alright otherwise.

Regarding the calling of these reusable workflow, I would actually prefer the source field there to be an option and have our known sources listed.
Even though that forces us to update it whenever we change any soure name in any way.

.github/workflows/build-app.yml
datenreisender: For the [build workflow](https://github.com/NordicSemiconductor/pc-nrfconnect-boilerplate/blob/add_artifactory_github_actions/.github/workflows/build.yml) the default is always correct. For the [release workflow](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/artifactory/github_actions/.github/workflows/release-app.yml) the default will be correct most of the time. So we only need to override occasionally.

.github/workflows/build-app.yml
datenreisender: It is used in the release workflow: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/artifactory/github_actions/.github/workflows/release-app.yml#L20

.github/workflows/release-app.yml
datenreisender: Here the reasoning is similar: When [releasing the latest commit to the `latest` source](https://github.com/NordicSemiconductor/pc-nrfconnect-boilerplate/blob/add_artifactory_github_actions/.github/workflows/release-latest.yml), the default ref is always correct. When [releasing manually to a specified source](https://github.com/NordicSemiconductor/pc-nrfconnect-boilerplate/blob/add_artifactory_github_actions/.github/workflows/release.yml) the default (latest commit on main) also seems like a good idea unless you explicitly want to specify something else.

.github/workflows/test-shared.yml
datenreisender: It is not really necessary, but I try to do as many steps towards releasing without actually releasing to check whether anything might go wrong _if_ we want to release.

datenreisender:
> Regarding the calling of these reusable workflow, I would actually prefer the source field there to be an option and have our known sources listed. Even though that forces us to update it whenever we change any soure name in any way.

I like that idea. It also means that I can switch from specifying the source name and access level separately to instead selecting them both together. 

This will make it easier (you only have to select one instead of two) and less error prone (because you cannot select wrong combinations like „official/internal“. 

So, I will implement something that we can define a workflow with an input like this:
```yaml
description: 'Source to release to'
type: choice
options:
    - official (external)
    - release-test (internal)
    - experimental (internal)

```

boundlesscalm (approved):

PR #977, opened 2025-02-17
Create a new source with script or GitHub action

datenreisender:
Part of [NCD-795](https://nordicsemi.atlassian.net/browse/NCD-795):

Add script and GitHub Actions workflow to create a source. 

The added description field is not yet used, but the idea is to already maintain it and eventually show it in the launcher.

### DIFF
--- a/.github/workflows/create-source.yml
+++ b/.github/workflows/create-source.yml
@@ -1,0 +1,47 @@
+name: Create source
+
+on:
+    workflow_dispatch:
+        inputs:
+            id:
+                description: ID (e.g. release-test)
+                required: true
+                type: string
+            name:
+                description: Name (e.g. Release Test)
+                required: true
+                type: string
+            access-level:
+                description: Access level (e.g. internal)
+                required: true
+                type: choice
+                options:
+                    - internal
+                    - internal-confidential
+                    - external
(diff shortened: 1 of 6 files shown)


boundlesscalm (approved):
While I'm a bit worried that we will forget the context this was created in and feel like just creating the source if it doesn't exist would be nicer, I do think it is better to have this as an explicitly separate action.
Makes it so that we have to do this consciously, and in the worst case it is also a fail-safe if we give it the wrong source to upload to.

datenreisender:
@boundlesscalm I also mentioned this action and script in the documentation about releasing an app, making it more likely to remember that we have it: https://nordicsemi.atlassian.net/wiki/spaces/WYLND/pages/1363739420/nRF+Connect+for+Desktop+App+Release+Procedure#Release-to-an-unforeseen-source

And even if we forgot this is here or why and e.g. we rather create a source manually by creating a `source.json` in the right place, I see little harm in having this action/script. 

Also, if I really do create a new UX for adding sources ([NCD-1188](https://nordicsemi.atlassian.net/browse/NCD-1188)) I probably will add new metadata on Artifactory to find all available sources, which will make this script even more valuable to use.

PR #980, opened 2025-02-21
feat: allow for the group heading to be a reactnode

boundlesscalm:
This is for npm to allow the usage of documentationtooltips

Sadly this specific case would override the interaction mouse pointer when hovering over the title
It would also conflict if the "title" is used so that has to be considered

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,12 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 201.0.0 - 2025-02-21
+
+### Changed
+
+-   `Group` heading is changed to `React.ReactNode` instead of `string`.
+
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -7,6 +7,12 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 201.0.0 - 2025-02-21
+
+### Changed
+
+-   `Group` heading is not `React.ReactNode` instead of `string`.
```
kylebonnici, line 14:
```suggestion
-   `Group` heading is a `React.ReactNode` instead of `string`.
```

kylebonnici (approved):

Changelog.md
boundlesscalm: Was meant to write `now`...
Thanks

PR #983, opened 2025-03-05
Migrate Notice Box from Quickstart to the shared

KievDevel:


src/Link/ExternalLink.tsx
```diff
@@ -16,6 +17,10 @@ export default ({ href, label = href }: { href: string; label?: string }) => (
         className={classNames(
             'tw-preflight tw-text-nordicBlue hover:tw-underline'
         )}
+        onClick={event => {
+            telemetry.sendEvent('Visiting link', { href });
+            event.stopPropagation();
+        }}
```
KievDevel, line 23: Based on the code if there is no telemetry it should be silently ignored, so I expect no regressions

boundlesscalm (approved):
LGTM

PR #985, opened 2025-03-05
Add title for the ExternalLink

KievDevel:


datenreisender (approved):
Setting the title, which adds a popup on hover, probably only makes sense if a `label` is set (because otherwise the `href` would also be the `label`). But I think it is fine to set the title in any case.

boundlesscalm (approved):
Would agree with what Marko mentioned

boundlesscalm (approved):

PR #993, opened 2025-04-08
feat: log onDeviceDisconnected, onDeviceConnected and onDeviceIsReady…

boundlesscalm:
… events

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,18 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 207.0.0 - 2025-04-08
+
+### Added
+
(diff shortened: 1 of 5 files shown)


Changelog.md
```diff
@@ -7,6 +7,18 @@ This project does _not_ adhere to
...
+
+### Added
+
+-   Logging `onDeviceDisconnected`, `onDeviceConnected` and `onDeviceIsReady`
+    events.
+
+### Steps to upgrade when using this package
+
+-   Remove logging of `onDeviceDisconnected`, `onDeviceConnected` and
+    `onDeviceIsReady` events.
```
greg-fer, line 20: We should not have such custom sections in the changelog :/ This should be just a continuation of the entry above.

boundlesscalm: Could you explain? This is a convention we have had for some time.
It's a convention we created for ourselves but it makes it very clear on whether and what steps are required when breaking changes occur.

PR #994, opened 2025-04-08
refactor: more consistent onDeviceIsReady log

boundlesscalm:
Added a duplicate log previously which I removed.

Also moved the success log message so it would be called in all cases

### DIFF
--- a/src/Device/deviceSetup.ts
+++ b/src/Device/deviceSetup.ts
@@ -83,6 +83,10 @@ export const prepareDevice =
     ): AppThunk<RootState, Promise<void>> =>
     async dispatch => {
         const onSuccessWrapper = (d: Device) => {
+            logger.info(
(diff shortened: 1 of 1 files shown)


src/Device/deviceSetup.ts
```diff
@@ -83,6 +83,10 @@ export const prepareDevice =
     ): AppThunk<RootState, Promise<void>> =>
     async dispatch => {
         const onSuccessWrapper = (d: Device) => {
+            logger.info(
+                `Connected to device with the serial number ${device.serialNumber} ` +
```
greg-fer, line 87: `the device`, but oh well, I was not invited to review :P

boundlesscalm: I'll add it in a separate PR no worries

PR #1000, opened 2025-05-05
Specify core version of nrfutil

datenreisender:
Apps must now specify what core version of nrfutil they depend on. This is to ensure that more of the environment the app runs in is locked for an app and future updates, in nrfutil do not break the app unexpectedly.

If an app e.g. specifies that it requires nrfutil core 8.0.0 and nrfutil device 2.9.0, then that sandbox will be created in the path `nrfutil-sandboxes/8.0.0/device/2.9.0`.

If the launcher encounters an old app, which does not specify the core version yet, it will default to using nrfutil core 8.0.0, but use the old path, because those apps expect it there, e.g. `nrfutil-sandboxes/device/2.9.0`.

(On macOS with Apple Silicon there would be an additional segment `arm64/` after `nrfutil-sandboxes/` in those examples)

This is used by https://github.com/NordicSemiconductor/pc-nrfconnect-launcher/pull/1125.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,13 +7,24 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 212.0.0 - 2025-06-05
+
+### Added
+
+-   Specify what core version of nrfutil an app depends on.
+
+### Steps to upgrade when using this package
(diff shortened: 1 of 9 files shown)


kylebonnici (commented):
Looks great 🥇

boundlesscalm (approved):
LGTM

PR #1003, opened 2025-05-09
Cleanup nrfutil

datenreisender:
Collection of cleanups to the nrfutil code that I came across while working on the code or reviewing other code but I didn't want to mix with other changes, to keep them clearer. Other changes should be merged first, esp. after we get the currently tested release of the launcher out.


nrfutil/device/program.ts
```diff
@@ -62,33 +62,41 @@ export const isNordicDfuProgrammingOptions = (
...
-            args.push(`chip_erase_mode=${options.chipEraseMode}`);
-        if (options.reset) args.push(`reset=${options.reset}`);
-        if (options.verify) args.push(`verify=${options.verify}`);
-    } else if (isMcuBootProgrammingOptions(options)) {
-        if (options.mcuEndState)
-            args.push(`mcu_end_state=${options.mcuEndState}`);
-        if (options.netCoreUploadDelay)
-            args.push(
+        return [
+            `chip_erase_mode=${options.chipEraseMode}`,
```
boundlesscalm, line 70: Why does this not have the boolean evaluator?

nrfutil/device/program.ts
```diff
@@ -62,33 +62,41 @@ export const isNordicDfuProgrammingOptions = (
...
+            options.mcuEndState && `mcu_end_state=${options.mcuEndState}`,
+            options.target && `target=${options.target}`,
+            options.netCoreUploadDelay &&
                 `net_core_upload_delay=${Math.round(
                     options.netCoreUploadDelay
-                )}`
-            );
-        if (options.target) args.push(`target=${options.target}`);
-    } else if (isNordicDfuProgrammingOptions(options)) {
-        if (options.mcuEndState)
```
boundlesscalm, line 98: Why not keep this check?

boundlesscalm (approved):
Some small comments but they are mostly just nits

nrfutil/sandbox.ts
```diff
@@ -88,31 +88,22 @@ const commonParser = <Result>(
 };
 
 export class NrfutilSandbox {
-    baseDir: string;
-    module: string;
-    version: string;
-    coreVersion: string | undefined; // Must only be undefined when the launcher creates a sandbox for a legacy app, which does not specify the required core version
     onLoggingHandlers: ((logging: LogMessage, pid?: number) => void)[] = [];
     logLevel: LogLevel = isDevelopment ? 'error' : 'off';
-    env: ReturnType<typeof this.prepareEnv>;
```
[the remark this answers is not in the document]
datenreisender: The `env` field is still there: https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/e64e9b4e6ff963272c023c497b1dd68663d35a50/nrfutil/sandbox.ts#L50

I removed the type because TypeScript figures out the exact same time by the assignment in the constructor.

nrfutil/device/batch.ts
```diff
@@ -275,7 +266,7 @@ export class Batch {
                 return tempFilePath;
             };
             const tempFilePath = saveTemp();
-            args = ['--firmware', tempFilePath].concat(args);
+            args = ['--firmware', tempFilePath, ...args];
```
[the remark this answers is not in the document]
datenreisender: Oversight. Thanks for pointing it out, changed that in aab28b7bd3dfa2a39df278596a8d6c4abd096aac. Also discovered that the second spread of `callbacks` below was unnecessary and replaced that.

nrfutil/device/program.ts
datenreisender: According to [the type predicate `isJLinkProgrammingOptions`](https://github.com/NordicSemiconductor/pc-nrfconnect-shared/blob/main/nrfutil/device/program.ts#L45-L48) `chip_erase_mode` must always be defined if the `options` are of type `JLinkProgrammingOptions`. I also changed the field in the type to be mandatory.

nrfutil/device/program.ts
datenreisender: Because there are only exactly three possibilities for what type the options can be. So it made the code simpler to remove it.

An alternative would have been to throw an exception in this case. But I saw no need for that and so decided to use the “less code” alternative.

PR #1010, opened 2025-06-17
Add script to zap the Artifactory cache

datenreisender:
Because our personal tokens do not have the permission to zap the cache (but the one in the GitHub Actions secret has it), add a GitHub Action for it.


.github/workflows/zap-cache.yml
```diff
@@ -0,0 +1,72 @@
...
+                          throw new Error('No valid paths provided. Paths must start with "external/ncd/" and contain only word characters, slashes, and hyphens.');
+                      }
+
+                      console.log('Paths to process:', paths);
+
+                      const success = [];
+                      const errors = [];
+
+                      for (const path of paths) {
+                          const url = `https://files.nordicsemi.cn/artifactory/swtools-cache/${path}`;
```
boundlesscalm, line 40: Maybe I misunderstood or am missing some context, but why only the .cn one?
Aren't there 2 caches?

datenreisender: At least currently there is only files.nordicsemi.cn. 

files.nordicsemi.com is the main server (which also goes by the name eu.files.nordicsemi.com).

PR #1030, opened 2025-07-11
Feat/persist groups collapse state

tsvetelinpetrov:


src/Group/Group.tsx
```diff
@@ -91,8 +96,22 @@ export const Group = ({
...
+                collapseStatePersistanceId
+            );
+            if (persistedState !== undefined) {
+                return persistedState;
+            }
+        }
+        return defaultCollapsed;
+    };
+
+    const [collapsed, setCollapsed] = useState(getInitialCollapseState());
```
datenreisender, line 114: I thought too late about this, but you might want to slightly change this line: 

```suggestion
    const [collapsed, setCollapsed] = useState(getInitialCollapseState);
```

By passing the function instead of the invoking it, the initialisation code is only run when needed: https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state

This is not _so_ relevant in this case, but it is a good habit to do it like this if it is possible with little effort.

PR #1033, opened 2025-07-16
Feat/limit number input chars

tsvetelinpetrov:


boundlesscalm (approved):
Looks good.
May not be important but could you share a picture of it in action in this PR?

PR #1037, opened 2025-07-17
Feat/add narrower dropdown variant

tsvetelinpetrov:
image (2)

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -79,7 +79,7 @@ export default <T,>({
                         : classNames(
                               'tw-bg-gray-700 tw-text-white',
                               size === 'sm'
-                                  ? 'tw-h-6 tw-pl-2 tw-pr-1 tw-text-[10px]'
+                                  ? 'tw-h-6 tw-pl-2 tw-pr-1 tw-text-2xs'
                                   : 'tw-h-8 tw-px-2'
                           )
                 )}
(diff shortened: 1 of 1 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -116,7 +124,10 @@ export default <T,>({
                 {items.map(item => (
                     <button
                         type="button"
-                        className="tw-bg-transparent tw-clear-both tw-block tw-h-6 tw-w-full tw-whitespace-nowrap tw-border-0 tw-px-2 tw-py-1 tw-text-left tw-font-normal tw-text-white hover:tw-bg-gray-600 focus:tw-bg-gray-600"
+                        className={classNames(
+                            'tw-bg-transparent tw-clear-both tw-block tw-h-6 tw-w-full tw-whitespace-nowrap tw-border-0 tw-px-2 tw-py-1 tw-text-left tw-font-normal tw-text-white hover:tw-bg-gray-600 focus:tw-bg-gray-600',
+                            size === 'sm' && 'tw-text-[10px]'
```
kylebonnici, line 129: I am find with 'tw-text-[10px]' but if there is an equivalent like 'tw-text-sm' or 'tw-text-xs' defined that case size 10px I would recommend that

boundlesscalm (approved):

src/Dropdown/Dropdown.tsx
```diff
@@ -116,7 +124,10 @@ export default <T,>({
                 {items.map(item => (
                     <button
                         type="button"
-                        className="tw-bg-transparent tw-clear-both tw-block tw-h-6 tw-w-full tw-whitespace-nowrap tw-border-0 tw-px-2 tw-py-1 tw-text-left tw-font-normal tw-text-white hover:tw-bg-gray-600 focus:tw-bg-gray-600"
+                        className={classNames(
```
kylebonnici, line 127: Given that the dropdown items are also smaller you also need to look at 
`
  data-height={
                    numItemsBeforeScroll > 0 &&
                    items.length > numItemsBeforeScroll
                }
`

to make sure that the correct about of item are shown when in sm config

kylebonnici (changes requested):

src/Dropdown/Dropdown.tsx
tsvetelinpetrov: Added a custom tailwind class: tw-text-2xs

https://github.com/NordicSemiconductor/pc-nrfconnect-shared/pull/1038

kylebonnici (approved):
After discussing it was conformed the the dropdown item high remains 24px hence no changes are needed

PR #1050, opened 2025-08-01
doc: changed JLink to SEGGER J-Link

greg-fer:
Also changed nrfutil to nRF Util when referring to the entire tool. Changed `nrfutil-<command>` to `nrfutil <command>`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -115,7 +115,7 @@ every new version is a new major version.
 
 ### Added
 
--   Specify what core version of nrfutil an app depends on.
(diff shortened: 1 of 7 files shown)


datenreisender (approved):
This also changes some internal IDs like `No J-Link installed`. 

Because they are internal and never presented to users there was no real need to change them, but it is ok. 

Because the IDs are also used in the launcher, we will need to update the launcher but that is also ok. This will show up as type check errors when trying to build the launcher, so there is no risk in missing this.

PR #1057, opened 2025-09-02
shared: ctrl+r in shortcuts

greg-fer:
Added entry for ctrl+r in shortcuts for all apps.
NCD-696.

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -10,8 +10,8 @@ import React, { FC, ReactNode, useEffect, useMemo, useRef } from 'react';
 import Carousel from 'react-bootstrap/Carousel';
 import { createRoot } from 'react-dom/client';
 import { useDispatch, useSelector } from 'react-redux';
-import { Reducer } from 'redux';
 import { getCurrentWindow } from '@electron/remote';
+import { Reducer } from 'redux';
 
 import { inMain as open } from '../../ipc/open';
 import { setNrfutilLogger } from '../../nrfutil/nrfutilLogger';


datenreisender:
I don't think the shortcut ctrl-r (cmd-r on Mac) is relevant for most users. Is there a reason to add it to the list of shortcuts? (I do not strongly oppose it, I am just curious and would not like to add all shortcuts to the shortcuts pane, because e.g. toggling the dev tools is even less relevant for most people.)

datenreisender:
BTW, while on the Mac NCD a menu bar is always visible (which also shows the shortcut for Restart Windows) there also is a menu bar on Windows and Linux, it just is hidden by default and one can make it visible by pressing and releasing “alt”.

src/App/App.tsx
```diff
@@ -108,6 +109,13 @@ const ConnectedApp: FC<ConnectedAppProps> = ({
         action: open.openLauncher,
     });
 
+    useHotKey({
+        hotKey: 'ctrl+r',
+        title: 'Restart app',
+        isGlobal: true,
+        action: () => getCurrentWindow().reload(),
```
datenreisender, line 116: To make this work the same as the current ctrl/cmd-r shortcut, this should be changed:
```suggestion
        action: () => getCurrentWindow().emit('restart-window'),
```
(The behaviour seems mostly the same but there seems to be some difference)

greg-fer:
> I don't think the shortcut ctrl-r (cmd-r on Mac) is relevant for most users. Is there a reason to add it to the list of shortcuts? (I do not strongly oppose it, I am just curious and would not like to add all shortcuts to the shortcuts pane, because e.g. toggling the dev tools is even less relevant for most people.)

There are some cases in docs ([example](https://docs.nordicsemi.com/bundle/nrf-connect-programmer/page/troubleshooting.html#application-returns-an-error)) where we mention this shortcut, so I thought it would be good to have it "officially" mentioned in the app.

greg-fer:
Agreed with @datenreisender to close this PR. The Ctrl/Cmd+R is a common OS shortcut, which means it is not app-specific. Docs can still mention it, but we don't need it as an entry in the shortcut list.

PR #1104, opened 2026-03-17
Upgraded to electron 41.0.2

maeldroem:
Upgraded to Electron 41 in order to have access to Popover API, Dialogs, position anchoring, etc.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,14 @@ every new version is a new major version.
 
 ### Steps to upgrade when using this package
 
+#### Electron 41
+
+Check the
+[Electron 41 changelog](https://www.electronjs.org/blog/electron-41-0).
+
+Moreover, the `engines.nrfconnect` entry in the `package.json` of apps must be
+at least `>=5.4.0`.
+
 #### Migrating the `Card` component
 
 The `title` and `titleButton` are no longer available on `Card`, and the overall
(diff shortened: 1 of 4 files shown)


maeldroem:
I agree with you all and i also want to point out that this PR can be delayed as much as needed, but some of my branches regarding the transition from bootstrap to tailwind will be based on this branch.

I'll classify this branch as a draft for the time being, unless you want to release it for 245 and set the launcher dependency of the apps to 244.

But if we wait a bit longer, once i'm done migrating shared components from bootstrap to tailwind i can look deeper in that upgrade to make sure nothing breaks.

As for the clipboard API deprecation, @datenreisender , from what i had checked there isn't any problem with it as we use `@electron/remote` which just makes main-thread-methods accessible to renderer processes without actually using the renderer process' API to do so.

datenreisender:
I just checked and an app built with using this, also requires a launcher built with using this. So, the “Steps to upgrade when using this package” section in this PR must also include an entry that the `engines.nrfconnect` field in `package.json` must be bumped. 

We have not decided yet what the version after the next of the launcher will be, but for now let us here just say that apps using this version of shared must bump `engines.nrfconnect` to at least `>=5.4.0`.

kylebonnici:
CC @datenreisender this can now be closed?

datenreisender:
> CC @datenreisender this can now be closed?

@kylebonnici Good catch. I close this.

PR #1121, opened 2026-04-24
[NCD-1629] Fix: Consistent card style

maeldroem:
Making the style consistent and definite until we have robust design guidelines.

### DIFF
--- a/src/About/About.tsx
+++ b/src/About/About.tsx
@@ -17,7 +17,7 @@ export interface AboutPaneProps {
 }
 
 export default ({ documentation, feedbackCategories }: AboutPaneProps) => (
-    <section className="tw-preflight tw-flex tw-flex-row tw-flex-wrap tw-justify-center tw-gap-4 tw-pb-4">
+    <section className="tw-preflight tw-flex tw-flex-row tw-flex-wrap tw-justify-center tw-gap-2">
         <ApplicationCard className="tw-max-w-80 tw-flex-1 tw-basis-60" />
         <DeviceCard className="tw-max-w-80 tw-flex-1 tw-basis-60" />
         {documentation && documentation.length && (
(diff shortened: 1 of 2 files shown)


SlavaNordic:
Disagree with the change. This does simply change styles and does not make anything consistent.

We also had a clear feedback from the Ketil to keep Card styles as they were designed before and I support this. This change makes card less dynamic and boring. We should not change design during refactoring and if we do, all changes must be reviewed. Can't approve.

@boundlesscalm @datenreisender tagging you if you have anything to bring.

src/About/About.tsx
```diff
@@ -17,7 +17,7 @@ export interface AboutPaneProps {
 }
 
 export default ({ documentation, feedbackCategories }: AboutPaneProps) => (
-    <section className="tw-preflight tw-flex tw-flex-row tw-flex-wrap tw-justify-center tw-gap-4 tw-pb-4">
+    <section className="tw-preflight tw-flex tw-flex-row tw-flex-wrap tw-justify-center tw-gap-2">
```
boundlesscalm: I don't understand this change or how it would be more consistent.
I may be missing something though, could you explain?

boundlesscalm: Just to be clear, I'm not necessarily opposed. This is just a personal request for clarification.

maeldroem: This is just because the gap between cards used to be 8px instead of 16px — PR Name is obsolete

SlavaNordic:
@maeldroem can you please add screenshots, before and after, to compare the change? seems now only About pane cards are affected

maeldroem:
# Before

image

----------

# After

image

@SlavaNordic

boundlesscalm (approved):
Looks fine to me
Not sure if @SlavaNordic has anymore comments though

datenreisender (approved):
Switching to `tw-gap-2` seems right to me, since that reverts to the 8 px gap that we also had before.

SlavaNordic (approved):
Thanks for providing the screenshots. Decreasing the spacing between the cards successfully groups them together, making them feel like a unified and cohesive set of related elements.

PR #1125, opened 2026-05-15
Fix: Application does not reopen when resetting defaults setting when some close critica section is active

kylebonnici:
https://nordicsemi.atlassian.net/browse/NCD-1656

Demo of fix 

https://github.com/user-attachments/assets/23f83403-bd6b-4570-add0-8f95f063312a

I was unable to just disable the reset default button as this would end up always being disabled in PPK given that until data is save we will always gave some close critical section.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,13 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 248.0.0 - Unreleased
+
+### Fixed
+
+- Application does not reopen when resetting defaults setting when some close
+  critica section is active.
+
 ## 247.0.0 - 2026-04-23
 
 ### Changed
@@ -48,14 +55,12 @@ structure of `Card` changed. Follow this format:
 import { Card } from '@nordicsemiconductor/pc-nrfconnect-shared';
 
 <Card>
-  <Card.Header className="tw-flex tw-flex-row tw-justify-between">
-    <Card.Header.Title cardTitle="your title here" />
-    { /* your button here */ }
-  </Card.Header>
-  <Card.Body>
-    { /* your card content here */ }
-  </Card.Body>
-</Card>
+    <Card.Header className="tw-flex tw-flex-row tw-justify-between">
+        <Card.Header.Title cardTitle="your title here" />
+        {/* your button here */}
+    </Card.Header>
+    <Card.Body>{/* your card content here */}</Card.Body>
+</Card>;
 ```
 
 ## 244.0.0 - 2026-01-22
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,8 +11,8 @@ every new version is a new major version.
 
 ### Fixed
 
-- Application does not reopen when resetting defaults setting when some close
-  critica section is active.
+- In case app is executing important task application will properly reopen when
+  resetting defaults.
 
 ## 247.0.0 - 2026-04-23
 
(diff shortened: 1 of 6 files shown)


src/ConfirmBeforeClose/confirmBeforeCloseSlice.ts
```diff
@@ -65,6 +69,9 @@ const slice = createSlice({
         setShowCloseDialog(state, action: PayloadAction<boolean>) {
             state.showCloseDialog = action.payload;
         },
+        setActionOnAllCompelte(state, action: PayloadAction<OnCompleteAction>) {
```
tsvetelinpetrov: Type. Should be setActionOnAllComplete

src/FactoryReset/FactoryResetButton.tsx
```diff
@@ -32,11 +34,13 @@ const FactoryResetButton: FC<Props> = ({
     classNames,
     large = false,
 }) => {
+    const dispatch = useDispatch();
     const [showDialog, setShowDialog] = useState(false);
     useRef(); // showdialog
     const defaultResetFn = () => {
         store().clear();
         logger.info('Successfully restored defaults');
+        dispatch(setActionOnAllCompelte('reload'));
```
tsvetelinpetrov: Sets actionOnAllComplete to 'reload' before calling reload(). If the user cancels the confirm-before-close dialog, the state stays as 'reload' and later normal close confirm dialog will reload the app instead of closing it.

kylebonnici: Fair point will revert back to close in that case! good catch!

Changelog.md
```diff
@@ -7,6 +7,13 @@ This project does _not_ adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html) but contrary to it
 every new version is a new major version.
 
+## 248.0.0 - Unreleased
+
+### Fixed
+
+- Application does not reopen when resetting defaults setting when some close
+  critica section is active.
```
SlavaNordic:
```suggestion
- In case app is executing important task application will properly
  reopen when resetting defaults.
```
maybe smth like this. still not perfect wording

SlavaNordic (dismissed):

tsvetelinpetrov (dismissed):

src/ConfirmBeforeClose/confirmBeforeCloseSlice.ts
```diff
@@ -105,3 +116,12 @@ export const preventAppCloseUntilComplete =
...
     };
+
+export const onUserConfirmAll = (): AppThunk => (_, getState) => {
+    const actionType = getActionOnComplete(getState());
+    if (actionType === 'close') {
+        getCurrentWindow().close();
+    } else {
+        getCurrentWindow().reload();
+    }
+};
```
datenreisender: In apps we usually keep thunks in `effects.ts` files instead of mixing them together with slices. But I see that we have not started this practice in shared yet.

datenreisender (dismissed):

src/ConfirmBeforeClose/confirmBeforeCloseSlice.ts
kylebonnici: I happy to move it. Any recommendation where effects.ts should be at?

datenreisender: @kylebonnici Same folder, keep it in `src/ConfirmBeforeClose` because that is what this effect is about.

PR #1126, opened 2026-05-22
Feature/SBOM

chunfantasy:
This PR is to generate SBOM for apps.

Once the app is updated with this feature from Shared, the app will generate the sbom.json file in the dist folder when build.

Note, providing SBOM from nrfutil and it's subcommand is not part of this PR.


datenreisender:
- Please describe the change in `Changelog.md`.
- Can we make the output on build less noisy?

PR #1134, opened 2026-06-05
Export pseudo button

tsvetelinpetrov:
Need to use the pseudo button in nAF app (look at the screenshot). Open for better suggestions.

image


datenreisender (dismissed):
I imagine that element in the nAF app contains `div`s, so a normal button does not cut it, right?

kylebonnici (dismissed):

PR #1135, opened 2026-06-05
nRF92 family will always have modem trait

SlavaNordic:


kylebonnici (dismissed):
LGTM, I did not check that what `nrfutil device device-info --serial-number .... --json-pretty` returns .... but if deviceFamily matches good to go

PR #1145, opened 2026-07-17
Remove the problematic ansi-regex lib

tsvetelinpetrov:


datenreisender (dismissed):
Please add some comment like 
```md
### Fixed

- Running Jest was broken with shared 254.
```
to the `Changelog.md`. It is unlikely that people start using an outdated version of shared but _if_ they do (e.g. when triaging problems) it should be easy to find an indicator why that version broke the tests.

datenreisender (approved):

PR #1148, opened 2026-07-29
Feat/autoupdate

formalmia:
For use in: https://github.com/nordicsemi/pc-nrfconnect-launcher/pull/1311

### DIFF
--- a/ipc/schema/packageJson.ts
+++ b/ipc/schema/packageJson.ts
@@ -28,6 +28,7 @@ const nrfConnectForDesktop = z.object({
     nrfutil: nrfModules.optional(),
     nrfutilCore: semver,
     html: z.string(),
+    autoUpdate: z.boolean().optional(),
     fixedSize: z
         .object({
             width: z.number().int().positive(),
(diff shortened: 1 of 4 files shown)


ipc/schema/packageJson.ts
```diff
@@ -48,6 +48,7 @@ const packageJsonApp = packageJson.extend({
     homepage: z.string().url().optional(),
     devDependencies: recordOfOptionalStrings.optional(),
     displayName: z.string(),
+    autoUpdate: z.boolean().optional(),
```
datenreisender: It would be better to put `autoUpdate` in `nrfConnectForDesktop` instead of having it top-level in `package.json`.

datenreisender (changes requested):

ipc/schema/packageJson.ts
formalmia: That makes sense.
I felt that nrfConnectForDesktop was reserved for tools so I was unsure and just put it here (and then ended up forgetting that it was something to consider)

